PR #33, opened 2020-01-30
Reconfigure mark one for consumption by external projects

rmainseas:
- Removes the `mark-one` alias in favor of relative paths to module to allow the components to be bundled properly by consuming apps (full write-up described in the commit message of 626b7d4).
- Update package.json to make the library entry point and type info available to package consumers
- Add `postinstall` script to transpile the JSX when pulling the package in from GitHub.


rmainseas:
Thinking about this some more, the only concern I might have about updating the `examples` folder like I did in 826553a would mean that the examples aren't _quite_ examples that you can copy/paste any more. 

So, maybe 826553a should be reverted, and the `mark-one` alias put back just for the purposes of the styleguide? Open to suggestions on this though...

jonseitz:
@rmainseas Yeah, I think you've raised a good point about the pastability of the example code. Let's revert for the examples, keep the alias, and make sure we're vigilant about using relative imports in future code reviews

PR #34, opened 2020-01-31
Button Variants

natalynyu:
This PR includes the creation of 4 button variants: Primary, Secondary, Danger, and Borderless. This was done by extending the BasicButton component

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#34 ](https://github.com/seas-computing/course-planner/issues/125)


src/Button/BorderlessButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+  FunctionComponent, ReactElement, useContext,
+} from 'react';
+import styled, { ThemeContext } from 'styled-components';
+import { darken } from 'polished';
+import { BaseTheme } from '../Theme';
+import { StyledButton } from './StyledButton';
+import { ButtonProps } from './ButtonProps';
+
+const StyledPrimaryButton = styled(StyledButton)`
+  background: ${({ theme }): string => (theme.color.background.info.light)};
```
jonseitz: I think the idea behind the borderless button was that it would appear on screen as just the label text, e,g, for those cases where we have an edit/delete icon in a table. In that case, I think we'd want to have a `transparent` background instead, and then provide some other visual indication for the `:hover` and `:active` states.

src/Button/BorderlessButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+  color: ${({ theme }): string => (theme.color.text.light)};
```
jonseitz: If we do go with a transparent background, it would probably make sense for `color` to be a prop, so you can adjust for contrast.

src/Button/ButtonProps.tsx
```diff
@@ -0,0 +1,12 @@
+import { ReactNode, MouseEventHandler } from 'react';
+
+export interface ButtonProps {
```
jonseitz: This should be a `.ts` file, since it doesn't contain any jsx

src/Button/DangerButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { darken } from 'polished';
+import { BaseTheme } from '../Theme';
+import { StyledButton } from './StyledButton';
+import { ButtonProps } from './ButtonProps';
+
+const StyledPrimaryButton = styled(StyledButton)`
+  background: ${({ theme }): string => (theme.color.background.negative.dark)};
+  color: ${({ theme }): string => (theme.color.text.light)};
+  &:hover {
+    background: ${({ theme }): string => (darken(0.2, theme.color.background.negative.dark))};
```
jonseitz: as noted in seas-computing/course-planner#125, these should use the `.medium` and `.dark` values for the base and hover backgrounds, respectively

src/Button/PrimaryButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { BaseTheme } from '../Theme';
+import { StyledButton } from './StyledButton';
+import { ButtonProps } from './ButtonProps';
+
+const StyledPrimaryButton = styled(StyledButton)`
+  background: ${({ theme }): string => (theme.color.background.info.light)};
+  color: ${({ theme }): string => (theme.color.text.light)};
+  &:hover {
+    background: ${({ theme }): string => (darken(0.2, theme.color.background.info.light))};
+  }
```
jonseitz: as noted in seas-computing/course-planner#125, these should use the `.medium` and `.dark` values for the base and hover backgrounds, respectively

src/Button/__tests__/BorderlessButton.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+    clickSpy = spy();
+    ({ getByText } = render(
+      <BorderlessButton clickHandler={clickSpy}>
+        Click Me
+      </BorderlessButton>
+    ));
+  });
+  afterEach(function () {
+    clickSpy.resetHistory();
+    cleanup();
```
jonseitz: You shouldn't need to call `cleanup()` in each test, since it's being done in the `.mochainit.ts` file in the project root.

_This applies to other button tests as well._

src/Button/__tests__/BorderlessButton.test.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { render, fireEvent, cleanup } from 'test-utils';
+import { spy } from 'sinon';
+import assert from 'assert';
+import BorderlessButton from '../BorderlessButton';
+
+describe('Borderless Button', function () {
+  let getByText;
+  let clickSpy;
```
jonseitz: These should be typed (e.g. `let clickSpy: SinonSpy;`)

_This applies to other button tests as well._

src/Button/__tests__/BorderlessButton.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+  afterEach(function () {
+    clickSpy.resetHistory();
+    cleanup();
+  });
+  it('renders', function () {
+    getByText('Click Me');
+  });
+  it('calls the click handler when clicked', function () {
+    fireEvent.click(getByText('Click Me'));
+    assert.equal(clickSpy.callCount, 1);
```
jonseitz: We should (almost) always use `assert.strictEqual` instead of `assert.equal`, since it compares both the value and type. This is not a major risk here, but if `click.callCount` return `true`, it would still pass, even though we need to see the value be `1`;

_This applies to other button tests as well._

src/Button/ButtonProps.tsx
natalynyu: Fixed through d075208!

src/Button/__tests__/BorderlessButton.test.tsx
natalynyu: Thanks! Fixed through d6fad8a.

src/Button/__tests__/BorderlessButton.test.tsx
natalynyu: Fixed through d6fad8a

src/Button/__tests__/BorderlessButton.test.tsx
natalynyu: Fixed through d6fad8a

src/Button/__tests__/BorderlessButton.test.tsx
jonseitz: Ah. I forgot that there are some issues with running tests in `watch` mode related to this. I think I've found a possible solution (clearing the node `require` cache in the `afterEach` block). I'll submit that as a separate PR

src/Button/DangerButton.tsx
natalynyu: Fixed through 224e5ef

src/Button/PrimaryButton.tsx
natalynyu: Fixed through 224e5ef

src/Button/BorderlessButton.tsx
natalynyu: Thanks for your help! Fixed through 769e9fd

src/Button/BorderlessButton.tsx
natalynyu: Fixed through 769e9fd

natalynyu:
All buttons were exported through commit a36b12e. Tests to make sure that click handler isn't called when disabled is truthy on the button were added through 61e7644. Thanks!

src/Button/StyledButton.tsx
```diff
@@ -0,0 +1,15 @@
+import styled from 'styled-components';
+
+export const StyledButton = styled.button`
+  background: ${({ theme }): string => (theme.color.background.medium)};
+  border-color: ${({ theme }): string => (theme.color.background.medium)};
```
jonseitz: I missed this on the first round, but we should also set `border-size` and `border-style` properties in this component so that it'll be inherited by the other buttons. I loaded the styleguide in firefox and the default button styles are a little different there than in chrome, so setting those explicitly here will make it more consistent across browsers.

src/Button/BorderlessButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+  DARK = 'dark',
+}
+
+const StyledBorderlessButton = styled(StyledButton)`
+  background: transparent;
+  border-style: none;
+  color: ${({ theme, color }): string => (theme.color.text[color])};
+  &:hover {
+    background: transparent;
+    color: ${({ theme, color }): string => (theme.color.text[color])};
```
jonseitz: I think we need to specify some kind of visual cue for the hover state to help users realize this is a button. Not 100% sure that would look like though. Can you do some research into how other frameworks/etc. handle that kind of UX? Or run it by Vittorio to see if he has ideas?

src/Button/__tests__/BasicButton.test.tsx
```diff
@@ -1,29 +1,37 @@
 import React from 'react';
-import { render, fireEvent, cleanup } from 'test-utils';
-import { spy } from 'sinon';
+import { render, fireEvent } from 'test-utils';
+import { spy, SinonSpy } from 'sinon';
 import assert from 'assert';
 import BasicButton from '../BasicButton';
 
 describe('Basic Button', function () {
   let getByText;
```
jonseitz: `getByText` should also have a type declaration in all of these tests.

src/Button/__tests__/BasicButton.test.tsx
```diff
@@ -1,29 +1,37 @@
...
-  let clickSpy;
+  let clickSpy: SinonSpy;
   beforeEach(function () {
     clickSpy = spy();
     ({ getByText } = render(
-      <BasicButton clickHandler={clickSpy}>
-        Click You
-      </BasicButton>
+      <>
```
jonseitz: For unit testing, we should only render and test only thing at a time, especially for things like there where the thing being tested (the `disabled` prop) causes a change in the behavior of the component. In most cases I'd suggest using `context` blocks to set up a separate rendered component where the variable under test is being changed. Doing that, the test suite could be structured as something like this psuedo-code:

```js
describe('BasicButton', function () {
...
  context('when disabled is true', function () {
    beforeEach(function () {
      render(<Button disabled>)
    });
    it('does not fire the click handler', function () {
      ...
    });
  });
  context('when disabled is false', function () {    
    beforeEach(function () {
      render(<Button>)
    });
    it('fires the click handler', function () {
      ...
    });
  });
```
The general idea is that unit tests should always try to isolate the individual _unit_ of code as much as possible. So here, we're rending two buttons side-by-side that call the same handler, and when we try to click the `disabled` button, that handler shouldn't be called. That's simple enough, but it does open the possibility that clicking the non-disabled button will affect the outcome of the test against the disabled one. That's highly unlikely in this this particular test scenario, but we should still always be following best practices. It's good discipline, for one, but also really helpful if we make future changes that increase the complexity of these components.

examples/Button/BasicButton.md
```diff
@@ -1,3 +1,3 @@
 ```jsx
-<BasicButton clickHandler={function(){alert('you clicked the button')}}>Click Me</BasicButton>
+<BasicButton clickHandler={function(){alert('You clicked the basic button')}}>Click Me!</BasicButton>
```
jonseitz: Can we reformat these examples to meet our eslint guidelines? (i.e. parens after function, line-widths, jsx props on separate lines, etc.) I know eslint won't actually flag these as problems, but I think we should try to make our examples as copy-and-pastable as possible.

src/Button/StyledButton.tsx
natalynyu: Oh true I should have included those!! Fixed through c5e3d7f.

src/Button/__tests__/BasicButton.test.tsx
natalynyu: @jonseitz I haven't been able to find a type for getByText. I tried to use the type ```GetByText```, but it gave me the error ``` Types of parameters 'text' and 'container' are incompatible``` so it looks like this is not the right type to use. 
I ended up declaring a type using the function declaration listed here: [https://testing-library.com/docs/dom-testing-library/api-queries#bytext](url)
Is the way that I did this through commit ddf4baf acceptable?

examples/Button/BasicButton.md
natalynyu: Fixed through e59d270!

src/Button/__tests__/BasicButton.test.tsx
natalynyu: Thanks for pointing this out!! That makes a lot of sense. Fixed through 3c1fde2.

src/Button/BorderlessButton.tsx
natalynyu: Ok looking into this

src/Button/__tests__/BasicButton.test.tsx
natalynyu: I ended up using `BoundFunction<GetByText>;` as the type for `getByText`

src/Button/BorderlessButton.tsx
natalynyu: We decided to add a `variant` prop and set the default text color of the BorderlessButton to `theme.color.background[variant].medium` and to `theme.color.background[variant].dark` on hover.

src/Theme/MarkOneTheme.ts
```diff
@@ -39,6 +39,20 @@ enum WEIGHT {
...
+  INFO = 'info',
+  POSITIVE = 'positive',
+  PRIMARY = 'positive',
+  NEGATIVE = 'negative',
+  SECONDARY = 'base',
+  DANGER = 'negative',
+  DEFAULT = 'base',
+  LIGHT = 'light',
+  DARK = 'dark',
+}
```
rmainseas: Should this line have a semi-colon at the end as-per eslint?

natalynyu: @rmainseas When I add a semi-colon there, eslint says that it's an unnecessary semi-colon

rmainseas: Interesting :thinking: Oh well, fair enough.

examples/Buttons/BorderlessButton.md
```diff
@@ -0,0 +1,16 @@
...
+  <BorderlessButton
+    onClick={function() {
+      alert('You clicked the borderless button')
+    }}
+    variant={VARIANT.BASE}
+  >
+    <FontAwesomeIcon icon={faArrowUp} size="small" />
+  </BorderlessButton>
+</div>
+```
```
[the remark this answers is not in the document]
natalynyu: Oh good idea! Added through 4d56821

src/Buttons/Button.tsx
```diff
@@ -0,0 +1,70 @@
...
+  disabled?: boolean;
+  /** Allows you to pass in style properties for the element */
+  theme?: object;
+  /** Allows you to pass in a variant property from the VARIANT enum */
+  variant: VARIANT;
+}
+
+const StyledButton = styled.button<ButtonProps>`
+  background: ${({ theme, variant }): string => (theme.color.background[variant].medium)};
+  border-color: ${({ theme, variant }): string => (theme.color.background[variant].medium)};
```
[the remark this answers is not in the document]
natalynyu: @jonseitz I'm a bit confused about what color to use for the border color, because originally I used the gray border for all my buttons, and then Chris asked me to change the border color to a color slightly darker than the button color (but if the color is `medium` and `dark` on hover, then we would need a `darker` option for the border on hover). Then in Vittorio's latest screenshot (below) he has the same border color as the button except for the basic button, but he also noted that the color contrast between the button text and color is enough that we don't need a darker border. Should I change the border to `theme.border.light` for now and then we can discuss in the course planning meeting? 
image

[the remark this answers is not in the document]
natalynyu: Removed borders in response to our discussion through commit 4b66bd2

src/Buttons/Button.tsx
```diff
@@ -24,17 +24,13 @@ export interface ButtonProps {
 
 const StyledButton = styled.button<ButtonProps>`
   background: ${({ theme, variant }): string => (theme.color.background[variant].medium)};
-  border-color: ${({ theme, variant }): string => (theme.color.background[variant].medium)};
-  border-style: solid;
-  border-width: 1px;
```
jonseitz: We need to add `border: none` to remove the default browser button border:

natalynyu: Ah, I must be going blind! Fixed through 20d8b7d

PR #35, opened 2020-02-07
Set up automated publishing to npm

jonseitz:
This PR includes multiple steps necessary to start publishing builds of `mark-one` on NPM:

1. Only includes the necessary files in the bundled version (e.g., remove tests, config files, etc.)
2. Run the compilation before publishing automatically
  - Can be tested with `npm publish --dry-run`
3. Add travis scripts to:
    1. On an untagged commit to develop or master, bump the version number and commit back to github
    2. On a tagged commit, build the package and deploy to npm, tagging develop builds as `mark-one@dev`

This double build is necessary because NPM only lets you publish unique versions, and we need to make sure those version changes are persisted in GitHub. I'm open to suggestions on other ways to do that.

fixes seas-computing/course-planner#142


.travis_version_bump.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+
+# Set the user name and email to match the API token holder
+git config --global user.email $DEV_TEAM_EMAIL 
+git config --global user.name "SEAS Computing Robot"
```
rmainseas: :robot:

PR #41, opened 2020-03-02
Update VARIANT import to have relative paths

natalynyu:
Instead of importing `VARIANT` enum directly from the `Theme` path, update the path to be relative to solve course-planner import error in which the `Button` and `BorderlessButton` components could not be found.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#117](https://github.com/seas-computing/course-planner/issues/117)


natalynyu:
> This looks fine, but we should also update the import lines in the test files for both of these as well.

Sorry, I should have thought of that - fixed through 376f27c.

PR #43, opened 2020-03-04
Table & TableCell Components: Remove Table margin and add TableCell alignment prop

natalynyu:
This PR removes the margin around the `Table` component, since the margin was causing a horizontal scrollbar to appear in Course Planning. The margin was moved to the `PageBody` component instead. Also, the prop `alignment` was added to the `TableCell` component so that users can adjust how the content within the `td` is positioned (`text-align` is set to `left`, `right`, or `center`). The default value of the `alignment` prop is `left`.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#117](https://github.com/seas-computing/course-planner/issues/117)

### DIFF
--- a/src/Tables/TableCell.tsx
+++ b/src/Tables/TableCell.tsx
@@ -17,7 +17,7 @@ export interface TableCellProps {
   /** Text or components to be displayed in the cell */
   children: ReactNode;
   /** Allows you to pass in a alignment property from the ALIGN enum */
-  alignment: ALIGN;
+  alignment?: ALIGN;
 }
 
 const StyledCell = styled.td<TableCellProps>`
--- a/src/Tables/TableCell.tsx
+++ b/src/Tables/TableCell.tsx
@@ -1,30 +1,49 @@
 import React, {
-  FunctionComponent, ReactElement, ReactNode, useContext,
+  FunctionComponent,
+  ReactElement,
+  ReactNode,
+  useContext,
 } from 'react';
 import styled, { ThemeContext } from 'styled-components';
 import { BaseTheme } from '../Theme';
 
+export enum ALIGN {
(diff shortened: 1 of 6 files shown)


src/Tables/TableCell.tsx
```diff
@@ -1,30 +1,49 @@
...
-  FunctionComponent, ReactElement, ReactNode, useContext,
+  FunctionComponent,
+  ReactElement,
+  ReactNode,
+  useContext,
 } from 'react';
 import styled, { ThemeContext } from 'styled-components';
 import { BaseTheme } from '../Theme';
 
+export enum ALIGN {
```
rmainseas: This could probably do with some jsdoc documentation on it

src/Tables/__tests__/Table.test.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { render } from 'test-utils';
 import convert from 'color-convert';
 import { strictEqual } from 'assert';
-import TableCell from '../TableCell';
+import TableCell, { ALIGN } from '../TableCell';
```
rmainseas: I would add some tests in here to make sure that:

- `ALIGN.LEFT`, `ALIGN.CENTER` and `ALIGN.RIGHT` do what you expect
- The default value is `ALIGN.LEFT`

natalynyu: Added tests through 5619e56

src/Tables/TableCell.tsx
natalynyu: Added jsdoc comment through 7434ce7

PR #45, opened 2020-03-05
Page heading component

rmainseas:
Create level 1 heading component (analogous to `<h1>`) to be used for page headers, application titles etc.

### DIFF
--- a/examples/Headings/PageTitle.md
+++ b/examples/Headings/PageTitle.md
@@ -1,4 +1,4 @@
 A top level heading, used for page headers, application titles etc.
 ```jsx
-<HeadingOne>Hello World</HeadingOne>
+<PageTitle>Hello World</PageTitle>
 ```
--- a/examples/Headings/PageTitle.md
+++ b/examples/Headings/PageTitle.md
@@ -1,4 +1,9 @@
 A top level heading, used for page headers, application titles etc.
+
+**Note:** For accessibility reasons, any given page should only ever have one(1)
+`<PageTitle>` component.
+
 ```jsx
+import { PageTitle } from 'mark-one';
 <PageTitle>Hello World</PageTitle>
 ```
--- a/src/Headings/HeadingOne.tsx
+++ b/src/Headings/HeadingOne.tsx
@@ -10,6 +10,7 @@ const Heading = styled.h1`
   font-family: ${({ theme }): string => theme.font.heading.family};
   font-size: ${({ theme }): string => theme.font.heading.size};
   font-weight: ${({ theme }): string => theme.font.heading.weight};
+  color: ${({ theme }): string => theme.font.heading.color};
 `;
 
 const HeadingOne: FunctionComponent = (props): ReactElement => {
--- a/src/Headings/HeadingOne.tsx
+++ b/src/Headings/HeadingOne.tsx
@@ -1,0 +1,33 @@
+import React, {
+  FunctionComponent,
+  ReactElement,
+  useContext,
+} from 'react';
+import styled, { ThemeContext } from 'styled-components';
+import { BaseTheme } from '../Theme';
+
+const Heading = styled.h1`
+  font-family: ${({ theme }): string => theme.font.heading.family};
+  font-size: ${({ theme }): string => theme.font.heading.size};
+  font-weight: ${({ theme }): string => theme.font.heading.weight};
+`;
+
+const HeadingOne: FunctionComponent = (props): ReactElement => {
+  const {
+    children,
+  } = props;
+  const theme: BaseTheme = useContext(ThemeContext);
+  return (
+    <Heading
+      theme={theme}
+    >
+      { children }
+    </Heading>
+  );
+};
+
+HeadingOne.defaultProps = {
+  children: '',
+};
+
+export default HeadingOne;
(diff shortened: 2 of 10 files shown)


src/Headings/HeadingOne.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { BaseTheme } from '../Theme';
+
+const Heading = styled.h1`
+  font-family: ${({ theme }): string => theme.font.heading.family};
+  font-size: ${({ theme }): string => theme.font.heading.size};
+  font-weight: ${({ theme }): string => theme.font.heading.weight};
+  color: ${({ theme }): string => theme.font.heading.color};
+`;
+
+const HeadingOne: FunctionComponent = (props): ReactElement => {
```
jonseitz: `FunctionComponent` should take an argument describing the props for the component (see other components in the project for examples).

src/Headings/HeadingOne.tsx
```diff
@@ -0,0 +1,34 @@
...
+import styled, { ThemeContext } from 'styled-components';
+import { BaseTheme } from '../Theme';
+
+const Heading = styled.h1`
+  font-family: ${({ theme }): string => theme.font.heading.family};
+  font-size: ${({ theme }): string => theme.font.heading.size};
+  font-weight: ${({ theme }): string => theme.font.heading.weight};
+  color: ${({ theme }): string => theme.font.heading.color};
+`;
+
```
jonseitz: Missing documentation for the component itself and its props

src/Headings/HeadingOne.tsx
```diff
@@ -0,0 +1,34 @@
...
+  useContext,
+} from 'react';
+import styled, { ThemeContext } from 'styled-components';
+import { BaseTheme } from '../Theme';
+
+const Heading = styled.h1`
+  font-family: ${({ theme }): string => theme.font.heading.family};
+  font-size: ${({ theme }): string => theme.font.heading.size};
+  font-weight: ${({ theme }): string => theme.font.heading.weight};
+  color: ${({ theme }): string => theme.font.heading.color};
```
jonseitz: These should use the `theme.font.title` attributes, since the `h1` element should only be used for the top page title.

src/Headings/HeadingOne.tsx
rmainseas: Does that still apply even if there are no props (other than `children`?

src/Headings/HeadingOne.tsx
rmainseas: Fixed in 40f0bd7

src/Headings/HeadingOne.tsx
rmainseas: I added documentation for props in 76e66fd. There is an example in the `examples` folder like the other components.

src/Headings/HeadingOne.tsx
rmainseas: Fixed in 76e66fd

jonseitz: I see this is resolved, but just to clarify: We should still explicitly specify that `children` is the only prop expected, as a signal (for `tsserver`, docs, etc,) that there are no other things that can be added.

src/Headings/HeadingOne.tsx
jonseitz: Similar to the note on the props, we should also be providing class-level doc blocks that can be picked up by `tsserver`. The examples are great for reference, but I think we need to include that API-style documentation wherever possible. So I think there is still work to be done here adding a general description for the component in a block above the declaration.

examples/Headings/PageTitle.md
```diff
@@ -0,0 +1,4 @@
+A top level heading, used for page headers, application titles etc.
```
jonseitz: I think we need a note here specifying that there should only be one `<PageTitle>` element on any "page" within the app...

rmainseas: Good idea.

rmainseas: Addressed in 4d138aac3fe5c71f72a5acefe9f243d938445a86

rmainseas:
I've also updated the `Header` documentation as you requested to use `PageTitle` instead.
eda8013c825308498dc9e0c010f64da84051a32b

src/Headings/HeadingOne.tsx
rmainseas: That makes sense, though I don't see them being used elsewhere in `mark-one`...

jonseitz: - [`src/Theme/MarkOneWrapper.tsx`](https://github.com/seas-computing/mark-one/blob/develop/src/Theme/MarkOneWrapper.tsx)
- [`src/Layout/Header.tsx`](https://github.com/seas-computing/mark-one/blob/develop/src/Layout/Header.tsx)
- [`src/Layout/Logo.tsx`](https://github.com/seas-computing/mark-one/blob/develop/src/Layout/Logo.tsx)

My feeling is that our documentation in the code should be as thorough and complete as possible. The styleguide is nice for actually being able to see how things will be rendered, but it's not a complete substitute for documentation.

rmainseas: Ah, I hadn't seen those. I was just looking at the other components such as [BorderlessButton](https://github.com/seas-computing/mark-one/blob/develop/src/Buttons/BorderlessButton.tsx) etc.

Should I add a ticket to get documentation into the other components?

jonseitz: Sure. We also need to make sure we're looking for stuff like this in code review.

rmainseas: Yeah. I updated that in 8aad5a9

PR #49, opened 2020-03-11
Adjust Header component to work better with PageTitle

rmainseas:
A small tweak to the `Header` component to make it work better with `PageTitle` by re-aligning child components using `align-items: baseline;`. This, in turn means that things are vertically center aligned, rather than aligned to the bottom of the header.

Before:  
image  
image

After:  
image  
image


rmainseas:
Ah. Good to know.

PR #53, opened 2020-04-03
Create Modal component

jonseitz:
Implements a full-screen overlay modal, suitable for use with edit screens, confirmation dialogs, and other communication purposes.

The `ModalHeader` and `ModalFooter` components will come in separate PRs, and those will be responsible for more of the look and feel of the modals. This component is mostly handling the appearing and disappearing effects.

### DIFF
--- a/src/Modals/__tests__/Modal.test.tsx
+++ b/src/Modals/__tests__/Modal.test.tsx
@@ -52,7 +52,6 @@ describe('Modal', function () {
           <div>
             <button
               type="button"
-              data-testid="open-button"
               onClick={() => { setModalOpen(!modalOpen); }}
             >
               Toggle Modal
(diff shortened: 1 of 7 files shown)


src/Modals/__tests__/Modal.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+        strictEqual(onCloseStub.callCount, 1);
+      });
+    });
+    context('When the modal changes from open to closed', function () {
+      beforeEach(function () {
+        const { getByTestId } = render(<ModalOpener />);
+        fireEvent.click(getByTestId('open-button'));
+        onOpenStub.resetHistory();
+        onCloseStub.resetHistory();
+        fireEvent.click(getByTestId('open-button'));
```
natalynyu, line 105: When I initially read this, I was confused by the use of `open-button`, because I thought we were trying to open the modal twice. Maybe it would be more clear if we called the test id `toggle-button`? What do you think?

jonseitz: Very good point. I actually just changed this to use `getByText` instead in 64df6231.

jonseitz:
Thanks for reviewing. We did talk about using the Esc key to clear the modal, but I think we decided to have that be handled by the parent component instead. I think you always want to be careful about adding/changing key bindings like that, and leaving it up to the implementation helps make sure that it's the "correct" behavior that the developer intended.

PR #54, opened 2020-04-07
Modal Footer

jonseitz:
Adds a Modal Footer component to the library, and integrates it into the Examples for the modal. Since the `ModalFooter` is just a `styled-component` wrapped in `withTheme`, I haven't added any tests for the component -- any tests would just be verifying that third-party libraries do what they're supposed to, and I don't think that's necessary for our needs. Happy to discuss further if you feel differently.

### DIFF
--- a/examples/Modals/ModalFooter.md
+++ b/examples/Modals/ModalFooter.md
@@ -1,0 +1,40 @@
+This is a utility component for adding a footer row to the `<Modal>`. It's set up to use `flexbox`, and fill in buttons from right to left. For example:
+
+```jsx
+import { useState } from 'react';
+import { Button } from 'mark-one';
+import Modal from '../../src/Modals/Modal.tsx';
+
+const TestModal = () => {
+  const [modalVisible, setModalVisible] = useState(false);
+  return (
+    <div>
+      <Button onClick={() => {setModalVisible(true)}}>
+        Open Modal
+      </Button>
+      <Modal
+        closeHandler={() => {setModalVisible(false)}}
+        isVisible={modalVisible}
+      >
+        <div>
+          This can be any arbitrary content
+        </div>
+        <ModalFooter>
+          <Button onClick={() => setModalVisible(false)}>
+            One
+          </Button>
+          <Button onClick={() => setModalVisible(false)}>
+            Two
+          </Button>
+          <Button onClick={() => setModalVisible(false)}>
+            Three
+          </Button>
+        </ModalFooter>
+      </Modal>
+    </div>
+  );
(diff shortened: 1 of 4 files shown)


src/Modals/ModalFooter.tsx
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * the content of the Modal Footer
+   */
+  children: ReactNode;
+}
+
+
+const styledModalFooter = styled.div<ModalFooterProps>`
+  display: flex;
+  flex-direction: reverse-row;
```
natalynyu, line 14: Do you mean `row-reverse` here? I'm not seeing a `reverse-row` option:
https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction
Also, could you clarify why we are ordering in reverse?

natalynyu: From the example, it seems like we are ordering in `row` order

examples/Modals/ModalFooter.md
```diff
@@ -0,0 +1,40 @@
+This is a utility component for adding a footer row to the `<Modal>`. It's set up to use `flexbox`, and fill in buttons from right to left. For example:
```
natalynyu, line 1: Should this be 'left to right'? If my comment from line 14 is wrong, you can disregard this

src/Modals/ModalFooter.tsx
jonseitz: Totally right, that should be `row-reverse` and I just did it wrong.

The reason for ordering in reverse is to follow the style of [Vittorio's mockups](https://github.com/seas-computing/course-planner/issues/169#issuecomment-601413769), where:
- **With only one button** the button is located in the bottom right corner
- **With two buttons** they're in the bottom right and bottom left corners (i.e., not right next to each other).
- I don't think we've specified what to do with 2+ buttons, but as of right now they'd be distributed evenly across the row.

The "reverse" part of it is less than ideal. We could use `justify-content: end;` with `flex-direction: row;`, but that would pack the items to the right, so the "cancel" button would be right next to the "confirm button" (not good). I'm open to other ideas though. Any thoughts?

examples/Modals/ModalFooter.md
jonseitz: rationale for "right to left" above

src/Modals/ModalFooter.tsx
natalynyu: Oh I see what you're saying with the `reverse` part of this now. I think this is the best way to go

PR #56, opened 2020-04-08
ModalBody Component

jonseitz:
This is a very small helper component that just provides padding to make the content of a modal line up with the header and footer.

### DIFF
--- a/src/Modals/ModalBody.tsx
+++ b/src/Modals/ModalBody.tsx
@@ -1,0 +1,26 @@
+import { ReactNode } from 'react';
+import styled, { withTheme } from 'styled-components';
+import { BaseTheme } from 'Theme';
+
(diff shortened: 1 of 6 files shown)


src/Modals/ModalBody.tsx
```diff
@@ -0,0 +1,26 @@
+import { ReactNode } from 'react';
+import styled, { withTheme } from 'styled-components';
+import { BaseTheme } from 'Theme';
+
+interface ModalBodyProps {
+  /**
+   * the content of the Modal Footer
```
natalynyu, line 7: Did you mean 'Modal Body'?

jonseitz: Yep, that's exactly right. fixed in 9eb9ebf

PR #58, opened 2020-05-06
Feature/201 spinner

farassadek:
Add spinner from fornawesome
issue: 
https://github.com/seas-computing/course-planner/issues/201

### DIFF
--- a/examples/Spinners/LoadSpinner.md
+++ b/examples/Spinners/LoadSpinner.md
@@ -5,7 +5,7 @@ A load spinner component that can be used to show the loading state of the data.
 </LoadSpinner>
 ```
 
-To style the text to have a dark light color.
+To style the text to have a light color. Note that a dark background is used to show the light text.
 ```jsx
 import { useContext } from 'react';
 import { ThemeContext } from 'styled-components';
--- a/examples/Spinners/LoadSpinner.md
+++ b/examples/Spinners/LoadSpinner.md
@@ -1,6 +1,5 @@
+A load spinner component that can be used to show the loading state of the data.
 ```jsx
-//import { LoadSpinner } from 'mark-one';
-
 <LoadSpinner>
   Loading Data
 </LoadSpinner>
--- a/examples/Spinners/LoadSpinner.md
+++ b/examples/Spinners/LoadSpinner.md
@@ -5,9 +5,9 @@ A load spinner component that can be used to show the loading state of the data.
 </LoadSpinner>
 ```
 
-To style the text to have a dark blue color instead of the default black color.
+To style the text to have a dark light color.
 ```jsx
-<LoadSpinner color='darkblue'>
+<LoadSpinner color='light'>
   Loading Data
 </LoadSpinner>
 ```
--- a/examples/Spinners/LoadSpinner.md
+++ b/examples/Spinners/LoadSpinner.md
@@ -1,0 +1,7 @@
+```jsx
+//import { LoadSpinner } from 'mark-one';
+
+<LoadSpinner>
+  Loading Data
+</LoadSpinner>
+```
--- a/examples/Spinners/LoadSpinner.md
+++ b/examples/Spinners/LoadSpinner.md
@@ -1,6 +1,4 @@
 A load spinner component that can be used to show the loading state of the data.
 ```jsx
-<LoadSpinner>
-  Loading Data
-</LoadSpinner>
+<LoadSpinner textMsg=' Loading Data ' />  
 ```
--- a/examples/Spinners/LoadSpinner.md
+++ b/examples/Spinners/LoadSpinner.md
@@ -7,7 +7,15 @@ A load spinner component that can be used to show the loading state of the data.
 
 To style the text to have a dark light color.
 ```jsx
-<LoadSpinner color='light'>
-  Loading Data
-</LoadSpinner>
+import { useContext } from 'react';
+import { ThemeContext } from 'styled-components';
+import { BaseTheme, SPINNER_TEXT } from 'mark-one';
+
+const theme = useContext(ThemeContext);
+
+<div style={{background: theme.color.background.darker}}>
+  <LoadSpinner textColor={SPINNER_TEXT.LIGHT}>
+    Loading Data
+  </LoadSpinner>
+</div>
 ```
--- a/examples/Spinners/LoadSpinner.md
+++ b/examples/Spinners/LoadSpinner.md
@@ -1,4 +1,13 @@
 A load spinner component that can be used to show the loading state of the data.
 ```jsx
-<LoadSpinner textMsg=' Loading Data ' />  
+<LoadSpinner>
+  Loading Data
+</LoadSpinner>
+```
+
+To style the text to have a dark blue color instead of the default black color.
+```jsx
+<LoadSpinner color='darkblue'>
+  Loading Data
+</LoadSpinner>
 ```
--- a/src/Spinners/LoadSpinner.tsx
+++ b/src/Spinners/LoadSpinner.tsx
@@ -11,7 +11,7 @@ export interface LoadSpinnerProps {
   /** Specifies the text under the spinner */
   children: string;
   /** Specifies the color of text under the spinner */
-  color?: SPINNER_TEXT;
+  textColor?: SPINNER_TEXT;
 }
 
 const StyledLoadSpinner = styled.div`
@@ -30,14 +30,14 @@ const LoadSpinner: FunctionComponent<LoadSpinnerProps> = (props):
 ReactElement => {
   const {
     children,
-    color,
+    textColor,
   } = props;
   const theme: BaseTheme = useContext(ThemeContext);
 
   return (
     <StyledLoadSpinner role="alert" theme={theme}>
       <FontAwesomeIcon icon={faCircleNotch} size="8x" spin />
-      <LoadSpinnerText color={color}>
+      <LoadSpinnerText color={textColor}>
         {children}
       </LoadSpinnerText>
     </StyledLoadSpinner>
--- a/src/Spinners/LoadSpinner.tsx
+++ b/src/Spinners/LoadSpinner.tsx
@@ -1,10 +1,10 @@
 import React, {
-  FunctionComponent, ReactElement, useContext
+  FunctionComponent, ReactElement, useContext,
 } from 'react';
 import styled, { ThemeContext } from 'styled-components';
-import { BaseTheme } from '../Theme';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+import { BaseTheme } from '../Theme';
 
 
 export interface LoadSpinnerProps {
@@ -24,14 +24,15 @@ const StyledLoadSpinner = styled.div`
   display: inline-block;
 `;
 
-const LoadSpinner: FunctionComponent<LoadSpinnerProps> = (props): ReactElement => {
+const LoadSpinner: FunctionComponent<LoadSpinnerProps> = (props):
+ReactElement => {
   const {
     children,
   } = props;
   const theme: BaseTheme = useContext(ThemeContext);
   return (
-    <StyledLoadSpinner theme={theme} >
-      <FontAwesomeIcon icon={faSpinner} spin/>
+    <StyledLoadSpinner theme={theme}>
+      <FontAwesomeIcon icon={faSpinner} spin />
       <div>
         { children }
       </div>
--- a/src/Spinners/LoadSpinner.tsx
+++ b/src/Spinners/LoadSpinner.tsx
@@ -1,10 +1,11 @@
 import React, {
-  FunctionComponent, ReactElement,
+  FunctionComponent, ReactElement, useContext,
 } from 'react';
-import styled from 'styled-components';
+import styled, { ThemeContext } from 'styled-components';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faCircleNotch } from '@fortawesome/free-solid-svg-icons';
 import LoadSpinnerText from './LoadSpinnerText';
+import { BaseTheme } from 'Theme';
 
 export interface LoadSpinnerProps {
   /** Specifies the text under the spinner */
@@ -15,11 +16,10 @@ export interface LoadSpinnerProps {
 
 const StyledLoadSpinner = styled.div`
   background: transparent;
-  color: lightblue;
+  color: ${({ theme }): string => (theme.color.background.info.medium)};
   display: inline-block;
   text-align:center;
   width:100%;
-  padding:5%;
 `;
 
 /**
@@ -31,8 +31,10 @@ ReactElement => {
     children,
     color,
   } = props;
+  const theme: BaseTheme = useContext(ThemeContext);
+
   return (
-    <StyledLoadSpinner role="alert">
+    <StyledLoadSpinner role="alert" theme={theme}>
       <FontAwesomeIcon icon={faCircleNotch} size="8x" spin />
       <LoadSpinnerText color={color}>
         {children}
--- a/src/Spinners/LoadSpinner.tsx
+++ b/src/Spinners/LoadSpinner.tsx
@@ -17,8 +17,9 @@ export interface LoadSpinnerProps {
 const StyledLoadSpinner = styled.div`
   background: transparent;
   color: ${({ theme }): string => (theme.color.background.info.medium)};
-  display: inline-block;
-  text-align:center;
+  display: flex;
+  flex-flow: column;
+  align-items: center;
   width:100%;
 `;
 
--- a/src/Spinners/LoadSpinner.tsx
+++ b/src/Spinners/LoadSpinner.tsx
@@ -6,24 +6,21 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faSpinner } from '@fortawesome/free-solid-svg-icons';
 import { BaseTheme } from '../Theme';
 
-
 export interface LoadSpinnerProps {
-  /** Allows you to pass in style properties for the element */
-  theme?: object;
   /** Specifies the text under the spinner */
   children: ReactElement;
 }
 
 const StyledLoadSpinner = styled.div`
   background: transparent;
-  color: ${({ theme }): string => (theme.color.background.dark)};
-  &:hover {
-    color: ${({ theme }): string => (theme.color.background.darker)};
-  }
+  color: ${({ theme }): string => (theme.color.background.darker)};
   font: ${({ theme }): string => (theme.font.heading)};
   display: inline-block;
 `;
 
+/**
+ * Data load spinner component with styled text under the spinner
+ */
 const LoadSpinner: FunctionComponent<LoadSpinnerProps> = (props):
 ReactElement => {
   const {
--- a/src/Spinners/LoadSpinner.tsx
+++ b/src/Spinners/LoadSpinner.tsx
@@ -1,0 +1,42 @@
+import React, {
+  FunctionComponent, ReactElement, useContext
+} from 'react';
+import styled, { ThemeContext } from 'styled-components';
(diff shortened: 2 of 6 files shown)


src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
...
+  children: ReactElement;
+}
+
+const StyledLoadSpinner = styled.div`
+  background: transparent;
+  color: ${({ theme }): string => (theme.color.background.dark)};
+  &:hover {
+    color: ${({ theme }): string => (theme.color.background.darker)};
+  }
+  font: ${({ theme }): string => (theme.font.heading)};
```
jonseitz, line 23: Unfortunately the font settings aren't this straightforward. You have to separately specif the font size, family, and weight with:

```suggestion
  font-family: ${({ theme }): string => (theme.font.heading.family)};
  font-size: ${({ theme }): string => (theme.font.heading.size)};
  font-weight: ${({ theme }): string => (theme.font.heading.weight)};
```

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
...
+export interface LoadSpinnerProps {
+  /** Allows you to pass in style properties for the element */
+  theme?: object;
+  /** Specifies the text under the spinner */
+  children: ReactElement;
+}
+
+const StyledLoadSpinner = styled.div`
+  background: transparent;
+  color: ${({ theme }): string => (theme.color.background.dark)};
```
jonseitz, line 19: We actually have a separate set of color values for text, and I think our `info` color is probably close to what the spinner in the older course-planner is, so this should be:

```suggestion
  color: ${({ theme }): string => (theme.color.text.info)};
```

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const LoadSpinner: FunctionComponent<LoadSpinnerProps> = (props):
+ReactElement => {
+  const {
+    children,
+  } = props;
+  const theme: BaseTheme = useContext(ThemeContext);
+  return (
+    <StyledLoadSpinner theme={theme}>
+      <FontAwesomeIcon icon={faSpinner} spin />
```
jonseitz, line 35: I think we might need to increase the size of the spinning icon a little bit. What about adding `size="3x"` to this?

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
...
+  background: transparent;
+  color: ${({ theme }): string => (theme.color.background.dark)};
+  &:hover {
+    color: ${({ theme }): string => (theme.color.background.darker)};
+  }
+  font: ${({ theme }): string => (theme.font.heading)};
+  display: inline-block;
+`;
+
+const LoadSpinner: FunctionComponent<LoadSpinnerProps> = (props):
```
jonseitz, line 27: We should have a JSdoc block here describing the component

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
...
+ReactElement => {
+  const {
+    children,
+  } = props;
+  const theme: BaseTheme = useContext(ThemeContext);
+  return (
+    <StyledLoadSpinner theme={theme}>
+      <FontAwesomeIcon icon={faSpinner} spin />
+      <div>
```
jonseitz, line 36: I think we also need a bit of space between the icon and the text. I think the best approach would probably be to define another styled-component like `StyledLoadSpinnerText` that has a `margin-top` property (we have some defined spacing in the theme, e.g. `theme.ws.medium`) and using that here instead of `<div>`. If you go that route, I'd suggest move the font properties from above to that component as well.

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
...
+`;
+
+const LoadSpinner: FunctionComponent<LoadSpinnerProps> = (props):
+ReactElement => {
+  const {
+    children,
+  } = props;
+  const theme: BaseTheme = useContext(ThemeContext);
+  return (
+    <StyledLoadSpinner theme={theme}>
```
jonseitz, line 34: Vittorio dropped [some links to accessibility guidelines](https://github.com/seas-computing/course-planner/issues/201#issuecomment-620164014) into the ticket for this. I _think_ the main thing is adding `role=alert` to this element, but you may want to double check with him.

src/Spinners/__tests__/LoadSpinner.test.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { render } from 'test-utils';
+import LoadSpinner from '../LoadSpinner';
+
+describe('Testing Data Loading Spinner', function () {
```
jonseitz, line 5: Super minor point, but our convention is for the top-level describe block to just be the name of the component (e.g. `LoadSpinner`). Just makes for cleaner test output.

src/Spinners/__tests__/LoadSpinner.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+    ));
+  });
+
+  it('renders', function () {
+    getByTestId('test-spinner');
+  });
+
+  it('renders value', function () {
+    getByText('Data Loading');
+  });
```
jonseitz, line 22: Both of these are really testing the same thing, so I'd suggest only using the `getByText` one since that's close to how a user would interact with this component.

That being said, lately I've been feeling that these kinds of tests ("renders what we put into it") aren't _actually_ all that useful. It might make sense to use the `container` property returned from render to do something like `container.querySelector('.fa-spinner.fa-spin')` and make sure that the icon is actually there. I don't know, I go back and forth over how much it's actually worth testing UI components like this...

examples/Spinners/LoadSpinner.md
```diff
@@ -0,0 +1,7 @@
+```jsx
```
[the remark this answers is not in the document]
farassadek: Ok

examples/Spinners/LoadSpinner.md
```diff
@@ -0,0 +1,7 @@
...
+//import { LoadSpinner } from 'mark-one';
```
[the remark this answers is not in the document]
farassadek: done

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
...
+} from 'react';
+import styled, { ThemeContext } from 'styled-components';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+import { BaseTheme } from '../Theme';
+
+
+export interface LoadSpinnerProps {
+  /** Allows you to pass in style properties for the element */
+  theme?: object;
```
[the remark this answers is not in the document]
farassadek: I fixed here but this need to be fixed in other places like the Icon component (Where I copied from).

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
+import React, {
+  FunctionComponent, ReactElement, useContext,
+} from 'react';
+import styled, { ThemeContext } from 'styled-components';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+import { BaseTheme } from '../Theme';
+
+
```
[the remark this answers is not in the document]
farassadek: ok

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
...
+  /** Specifies the text under the spinner */
+  children: ReactElement;
+}
+
+const StyledLoadSpinner = styled.div`
+  background: transparent;
+  color: ${({ theme }): string => (theme.color.background.dark)};
+  &:hover {
+    color: ${({ theme }): string => (theme.color.background.darker)};
+  }
```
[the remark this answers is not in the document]
farassadek: ok

src/Spinners/LoadSpinner.tsx
farassadek: ok

src/Spinners/__tests__/LoadSpinner.test.tsx
farassadek: ok

src/Spinners/__tests__/LoadSpinner.test.tsx
farassadek: I was thinking about the same exact thing, how much the test is useful and how deep I should go with the testing ,  I event started the container query selector and remove it. I guess we can add more tests as needed (if needed) .

src/Spinners/LoadSpinner.tsx
farassadek: done

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,44 @@
...
+  color?: string;
+}
+
+const StyledLoadSpinner = styled.div`
+  background: transparent;
+  color: lightblue;
+  display: inline-block;
+  text-align:center;
+  width:100%;
+  padding:5%;
```
jonseitz, line 22: I don't think we actually need a padding on this, since the parent component can handle it's spacing/placement.

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,44 @@
...
+export interface LoadSpinnerProps {
+  /** Specifies the text under the spinner */
+  children: string;
+  /** Specifies the color of text under the spinner */
+  color?: string;
+}
+
+const StyledLoadSpinner = styled.div`
+  background: transparent;
+  color: lightblue;
```
jonseitz, line 18: This should use one of our defined color constants, probably `theme.color.background.info.medium`

src/Spinners/LoadSpinner.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const StyledLoadSpinner = styled.div`
+  background: transparent;
+  color: ${({ theme }): string => (theme.color.background.dark)};
+  &:hover {
+    color: ${({ theme }): string => (theme.color.background.darker)};
+  }
+  font: ${({ theme }): string => (theme.font.heading)};
+  display: inline-block;
```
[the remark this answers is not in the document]
jonseitz: I don't think `inline-block` is correct here -- it should still be a `block` element.

I would also prefer to use [flexbox](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout) instead of `text-align` here, as noted in the suggested change above. They're almost doing the same thing, but flexbox is meant for handling more complex layouts like this, where `text-align` is specifically for handling the display of text.

src/Spinners/LoadSpinnerText.tsx
```diff
@@ -0,0 +1,41 @@
...
+  color?: string;
+}
+
+const StyledLoadSpinnerText = styled.div`
+  background: transparent;
+  color: ${({ color }): string => (color)};
+  font: ${({ theme }): string => (theme.font.heading)};
+  margin-top: ${({ theme }): string => (theme.ws.large)};
+  margin-bottom: ${({ theme }): string => (theme.ws.medium)};
+  display: block;
```
jonseitz, line 20: `div` are block elements by default, so this is unnecessary

src/Spinners/LoadSpinnerText.tsx
```diff
@@ -0,0 +1,41 @@
...
+  /** Specifies the color of the text under the spinner */
+  color?: string;
+}
+
+const StyledLoadSpinnerText = styled.div`
+  background: transparent;
+  color: ${({ color }): string => (color)};
+  font: ${({ theme }): string => (theme.font.heading)};
+  margin-top: ${({ theme }): string => (theme.ws.large)};
+  margin-bottom: ${({ theme }): string => (theme.ws.medium)};
```
jonseitz, line 19: I would leave out this `bottom-margin`, to provide more flexibility in how the spinner is displayed within the parent component.

src/Spinners/LoadSpinnerText.tsx
```diff
@@ -0,0 +1,41 @@
...
+export interface LoadSpinnerTextProps {
+  /** Specifies the text under the spinner */
+  children: string;
+  /** Specifies the color of the text under the spinner */
+  color?: string;
+}
+
+const StyledLoadSpinnerText = styled.div`
+  background: transparent;
+  color: ${({ color }): string => (color)};
```
jonseitz, line 16: Rather than allowing _any_ color value here, I'd prefer to provide a limited set of options with an `enum`, and to have those options correspond to the `light` and `dark` values in our theme. To set that up, you could do something like:

```tsx
export enum SPINNER_TEXT {
  LIGHT = 'light',
  DARK = 'dark',
}

export interface LoadSpinnerTextProps {
  /** Specifies the text under the spinner */
  children: string;
  /** Specifies the color of the text under the spinner */
  color?: SPINNER_TEXT;
}

const StyledLoadSpinnerText = styled.div<LoadSpinnerTextProps>`
  color: ${({ color, theme }): string => theme.color.text[color]};
  ...
`

StyledLoadSpinnerText.defaultProps = {
  color: SPINNER_TEXT.DARK;
}
```

The idea here being that we want to keep text colors consistent across the library (by using our only our defined colors), and we want to limit how much "damage" (in terms of inaccessible design) we let people do by only letting them use white or black text.

src/Spinners/LoadSpinnerText.tsx
```diff
@@ -0,0 +1,41 @@
...
+  font: ${({ theme }): string => (theme.font.heading)};
```
jonseitz, line 17: I noted this before, but I think it got lost in the refactoring:

You can't actually use the font values in our theme this way. You have to specific the family, size, and weight separately, with:

```tsx
   font-family: ${({ theme }): string => (theme.font.heading.family)};
   font-weight: ${({ theme }): string => (theme.font.heading.weight)};
   font-size: ${({ theme }): string => (theme.font.heading.size)};
```

src/Spinners/LoadSpinnerText.tsx
```diff
@@ -0,0 +1,41 @@
...
+const LoadSpinnerText: FunctionComponent<LoadSpinnerTextProps> = (props):
+ReactElement => {
+  const { children, color } = props;
+  const theme: BaseTheme = useContext(ThemeContext);
+  return (
+    <StyledLoadSpinnerText theme={theme} color={color}>
+      { children }
+    </StyledLoadSpinnerText>
+  );
+};
```
jonseitz, line 35: When we're directly mapping `props` and `children` to a styled component like this, it's much cleaner to just use:

```tsx
const LoadSpinnerText:
ForwardRefExoticComponent<LoadSpinnerTextProps> = withTheme(StyledLoadSpinnerText);

declare type LoadSpinnerText = ReactElement<LoadSpinnerProps>;

export default LoadSpinnerText;
```

(I've found that all the extra `type` annotations are needed to get the styleguide to properly recognize our components and their props)

src/Spinners/index.ts
```diff
@@ -0,0 +1 @@
+export { default as LoadSpinner } from './LoadSpinner';
```
jonseitz, line 1: If we go the route of using an `enum` for the color values above, that will also need to be exported from the library in here

jonseitz: Also, I just realized that we'll need to export these component from the library itself by adding `export * from './Spinners'` to the `src/index.ts` file.

src/Spinners/LoadSpinnerText.tsx
farassadek: make sense thought light color will be invisible on a white background

src/Spinners/LoadSpinnerText.tsx
farassadek: I did not know there is such way to define a component.

src/Spinners/index.ts
farassadek: I needed to export enum from the LoadSpinnerText text that I used in the LoadSpinner. Do I need to exported in src/Spinners/index.ts ?

examples/Spinners/LoadSpinner.md
```diff
@@ -0,0 +1,13 @@
...
+  Loading Data
+</LoadSpinner>
+```
+
+To style the text to have a dark light color.
+```jsx
+<LoadSpinner color='light'>
+  Loading Data
+</LoadSpinner>
+```
```
jonseitz, line 13: Here we want to use the actual enum value (e.g. `SPINNER_TEXT.LIGHT`) instead of the literal string `'light'`. It's effectively doing the same thing, but that's how we'd want these values to actually be used.

I would also recommend wrapping this spinner in a `div` with a dark background, as a way to show the different context in which is would be used.

````suggestion
```tsx
import { useContext } from 'react';
import { ThemeContext } from 'styled-components';
import { SPINNER_TEXT } from 'mark-one';

const theme: BaseTheme = useContext(ThemeContext);

<div style={{background: theme.color.background.darker}}>
  <LoadSpinner color={SPINNER_TEXT.LIGHT}>
    Loading Data
  </LoadSpinner>
</div>
```
````

File: src/Spinners/LoadSpinner.tsx [hunk 24]
```diff
@@ -0,0 +1,47 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCircleNotch } from '@fortawesome/free-solid-svg-icons';
+import { BaseTheme } from 'Theme';
+import LoadSpinnerText, { SPINNER_TEXT } from './LoadSpinnerText';
+
+export interface LoadSpinnerProps {
+  /** Specifies the text under the spinner */
+  children: string;
+  /** Specifies the color of text under the spinner */
+  color?: SPINNER_TEXT;
```
jonseitz, line 14: Can we rename this to `textColor`? and do so in the `LoadSpinnerText` component as well? I always find it better to be more specific with prop names.

src/Spinners/LoadSpinnerText.tsx
farassadek: Do you have documentation to help me understand how does your code work

jonseitz: OK, that's totally fine to leave this as is.

Just to give a little more explanation (which I should have done from the beginning, sorry): The [`withTheme` function](https://styled-components.com/docs/api#withtheme) is a higher-order component (a function that returns a new component) that's part of the the `styled-components` library that automatically injects the theme into the `theme` prop. It ends up generating the same output, but has the advantage of not requiring you to manually pass through props. In this case it's not a big deal, but there were some other components where we were passing though a number of different props and it was getting to be more of a mess.

src/Spinners/index.ts
jonseitz: This is correct here -- we need to export the `SPINNER_TEXT` enum, but not the `LoadSpinnerText`. This `src/Spinners/index.ts` is what gets exported from the main `src/index.ts`, so only these two exports will be available from the main application; the `LoadSpinnerText` itself will not be exported publicly.

src/Spinners/LoadSpinnerText.tsx
jonseitz: Oh, and as a bit of clarification on the type annotations:

- The `ForwardRefExoticComponent<LoadSpinnerTextProps>` is what's being returned from the `withTheme(StyledLoadSpinnerText)` call, and that needs to be in there to allow our styleguide tool to properly parse and associate the props with the component in the final output. 

- The `ReactElement<LoadSpinnerProps>` is a less-specific type, and declaring it that way lets us use this default export as a type itself. This is particularly useful with something like our `Table` components, where we're only allowing certain components to be used as `children`, though that's not really the case here.

src/Spinners/LoadSpinner.tsx
farassadek: yes , this make sense

jonseitz:
I approve this PR. Feel free to merge when ready!

PR #60, opened 2020-05-08
Update Table Components (Revision)

jonseitz:
I made a mistake in #59, leaving in imports from 'Theme' that should have been '../Theme'. I'm also changing the type for `rowSpan` and `colSpan` from `number` to `string`, since that's closer to how regular html would be written. 

With `number` the usage would have to be:

```tsx
<TableHeadingCell colSpan={2}>Heading</TableHeadingCell>
```

While with `string` the usage would be:

```tsx
<TableHeadingCell colSpan="2">Heading</TableHeadingCell>
```

### DIFF
--- a/src/Tables/Table.tsx
+++ b/src/Tables/Table.tsx
@@ -1,6 +1,6 @@
 import styled, { withTheme } from 'styled-components';
-import { BaseTheme } from 'Theme';
 import { ReactElement, ForwardRefExoticComponent } from 'react';
+import { BaseTheme } from '../Theme';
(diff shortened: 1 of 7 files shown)


jonseitz:
@rmainseas just highlighting this comment from commit 29c6917:

Upon further reflection, there's a good case for allowing both `number` and `string` here, for those times when the `colSpan` value is coming from a computed value like:

```tsx
<TableHeadingCell colSpan={columns.length}>Heading</TableHeadingCell>
```

PR #61, opened 2020-05-21
TableCellList components

jonseitz:
The current course-planner prototypes frequently use lists inside of a TableCell. I think it makes sense to have this as a special, Table specific design (as opposed to a more generalized List design), since this will probably be a pattern we employ in other circumstances.

This is needed for the work in seas-computing/course-planner#196

### DIFF
--- a/examples/Tables/Table.md
+++ b/examples/Tables/Table.md
@@ -178,7 +178,7 @@ import {
 
 Some tables may require more complex layouts, such as cells that span multiple columns/rows, or layered headers. For example:
 
-```tsx
+```jsx
 import {
   ALIGN,
   Table,
--- a/examples/Tables/Table.md
+++ b/examples/Tables/Table.md
@@ -176,7 +176,7 @@ import {
 
 ### More Complex Layouts
 
-Some tables may require more complex layouts, such as cells that span multiple columns/rows or layered, hierarchical headers. For example:
+Some tables may require more complex layouts, such as cells that span multiple columns/rows, or layered headers. For example:
 
 ```tsx
 import {
@@ -188,6 +188,8 @@ import {
   TableRowHeadingCell,
   TableHeadingCell,
   TableHeadingSpacer,
+  TableCellList,
+  TableCellListItem,
   TableHead,
 } from 'mark-one';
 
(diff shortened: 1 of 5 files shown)


examples/Tables/Table.md
```diff
@@ -176,7 +176,7 @@ import {
 
 ### More Complex Layouts
 
-Some tables may require more complex layouts, such as cells that span multiple columns/rows or layered, hierarchical headers. For example:
+Some tables may require more complex layouts, such as cells that span multiple columns/rows, or layered headers. For example:
 
 ```tsx
```
natalynyu, line 181: Is this supposed to be `tsx` or should it be `jsx`? The table wasn't showing up for me in the markdown until I changed it to `jsx`

examples/Tables/Table.md
```diff
@@ -232,33 +234,60 @@ import {
...
       <TableCell />
       <TableCell />
       <TableCell />
-      <TableCell>YES</TableCell>
+      <TableCell>
+        <TableCellList>
+          <TableCellListItem>
+            Michaels, Lianne
+          </TableCellListItem>
+        </TableCellList>
```
natalynyu, line 266: Just a question for my own understanding - Is the purpose of having a `<TableCellList>` with only one `<TableCellListItem>` that these are cells that would potentially have multiple list items in the future?

examples/Tables/Table.md
jonseitz: I think you need to be on the newest version of `styleguidist` for `tsx` to work (we bumped the version up in commit

examples/Tables/Table.md
jonseitz: I don't know if it's necessarily about "the future" so much as it is about consistency within the column, ie., showing that instructors _can_ have multiple entries, even if this specific cell only has one. If we're thinking in **types**, it's basically the difference between `Array<string>` and `string`.

I should also note that in practical usage, this would most likely be done by iterating through an array:

```tsx
<TableCellList>
  {instructors.map(
    ({ id, displayName }): ReactElement => (
      <TableCellListItem key={id}>
          {displayName}
      </TableCellListItem>
    )
  )}
</TableCellList>

```

natalynyu: Ok that makes sense

PR #63, opened 2020-06-11
Add 'id' property to Dropdown component

natalynyu:
The id property is needed to associate the label of the dropdown with the dropdown itself. I've added the id property to component in component definition, test component, and markdown example.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#214](https://github.com/seas-computing/course-planner/issues/214)

### DIFF
--- a/examples/Dropdown/Dropdown.md
+++ b/examples/Dropdown/Dropdown.md
@@ -1,13 +1,12 @@
 ```jsx
       import { useState } from 'react';
-      import { Label } from 'mark-one';
       function DropdownExample() {
         const [value, setValue] = useState('fall');
         return (
-          <Label
+          <label
             htmlFor="semesters"
-            label="Semester:"
           >
+          Semester:
           <Dropdown
           options={[
             {
@@ -28,7 +27,7 @@
             alert('You changed the selection to ' + event.target.value);
           }}
         />
-        </Label>
+        </label>
         );
       }
       <DropdownExample />
--- a/examples/Dropdown/Dropdown.md
+++ b/examples/Dropdown/Dropdown.md
@@ -1,8 +1,14 @@
 ```jsx
       import { useState } from 'react';
+      import { Label } from 'mark-one';
       function DropdownExample() {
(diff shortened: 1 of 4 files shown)


examples/Dropdown/Dropdown.md
```diff
@@ -15,6 +15,7 @@
             },
           ]}
           value={value}
+          id="semesters"
```
jonseitz, line 18: This example should be updated to include a Label

natalynyu: Ah, my mistake. I added this through 9bca048. 

And as for your notes above, good points! I'll make sure to include these notes in the ticket for improving the `Dropdown` component.

src/Forms/index.ts
```diff
@@ -1,2 +1,3 @@
 export { default as ValidationErrorMessage } from './ValidationErrorMessage';
 export { default as TextInput } from './TextInput';
+export { default as Label } from './Label';
```
jonseitz, line 3: One of our design decisions early on was that `Label` should not be exported from the library, since we don't want developers to have the options to use a `<Label>` on its own without associating it with a field.

examples/Dropdown/Dropdown.md
```diff
@@ -1,8 +1,14 @@
 ```jsx
       import { useState } from 'react';
+      import { Label } from 'mark-one';
```
jonseitz, line 3: This should use a relative path, or the example should just use a plain html `label` element until the `Dropdown` component can be refactored.

src/Forms/index.ts
natalynyu: Sorry, forgot about this! I removed export through commit a716905.

examples/Dropdown/Dropdown.md
natalynyu: Ok, I updated the example so that it uses the plain HTML `label` element through commit 8d0c344.

PR #64, opened 2020-06-15
Dropdown, TextInput, and Modal Components - Add Missing Props Needed for Faculty Admin Modals

natalynyu:
This PR adds missing properties needed to create the Faculty Admin modal.

`Dropdown`: `disabled` and `required` properties were added for form validation purposes.
`TextInput`: `required` property was added for form validation purposes.
`Modal`: `aria-labelledBy` property was added for accessibility purposes.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#214](https://github.com/seas-computing/course-planner/issues/214)

### DIFF
--- a/examples/Modals/Modal.md
+++ b/examples/Modals/Modal.md
@@ -20,7 +20,6 @@ const [modalVisible, setModalVisible] = useState(false);
     Open Modal
   </Button>
   <Modal
-    id="testModal"
     ariaLabelledBy="testButton"
     closeHandler={() => {setModalVisible(false)}}
     isVisible={modalVisible}
@@ -61,7 +60,6 @@ const [formValue, setFormValue] = useState('');
     Open Modal
   </Button>
   <Modal
-    id="testModal"
     ariaLabelledBy="testButton"
     onOpen={() => {setModalCount(modalCount + 1)}}
     onClose={() => { setFormValue('') }}
--- a/examples/Modals/Modal.md
+++ b/examples/Modals/Modal.md
@@ -16,7 +16,10 @@ import {
 const [modalVisible, setModalVisible] = useState(false);
 
 <div>
-  <Button onClick={() => {setModalVisible(true)}}>
+  <Button
+    id="testButton"
+    onClick={() => {setModalVisible(true)}}
+  >
     Open Modal
   </Button>
   <Modal
@@ -56,7 +59,10 @@ const [modalVisible, setModalVisible] = useState(false);
 const [formValue, setFormValue] = useState('');
 
 <div>
-  <Button onClick={() => {setModalVisible(true)}}>
+  <Button
+    id="testButton"
+    onClick={() => {setModalVisible(true)}}
+  >
     Open Modal
   </Button>
   <Modal
--- a/examples/Modals/Modal.md
+++ b/examples/Modals/Modal.md
@@ -20,6 +20,8 @@ const [modalVisible, setModalVisible] = useState(false);
     Open Modal
   </Button>
   <Modal
+    id="testModal"
+    ariaLabelledBy="testButton"
     closeHandler={() => {setModalVisible(false)}}
     isVisible={modalVisible}
   >
@@ -59,6 +61,8 @@ const [formValue, setFormValue] = useState('');
     Open Modal
   </Button>
   <Modal
+    id="testModal"
+    ariaLabelledBy="testButton"
     onOpen={() => {setModalCount(modalCount + 1)}}
     onClose={() => { setFormValue('') }}
     closeHandler={() => {setModalVisible(false)}}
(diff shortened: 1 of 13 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -14,12 +14,14 @@ export interface DropdownProps {
   onChange: ChangeEventHandler;
   /** The name of the dropdown */
   name: string;
-  /** An array of objects with the properties 'label' and 'value' for each dropdown option */
-  options: Array<{label: string; value: string}>;
+  /** An array of objects with the properties 'label', 'value', and 'disabled' for each dropdown option */
+  options: Array<{label: string; value: string; disabled?: boolean}>;
```
rmainseas, line 18: These additional options could do with having their own test cases. For example:

- What happens if I set `options.disabled` to `true`? 
- What happens if I set `options.disabled` to `false`
- What happens if I don't provide a value for `options.disabled`?

src/Dropdown/Dropdown.tsx
```diff
@@ -14,12 +14,14 @@ export interface DropdownProps {
...
   /** The currently selected dropdown value */
   value?: string;
   /** The dropdown value selected by default */
   defaultValue?: string;
+  /** If true, the dropdown is required to submit the form */
+  required?: boolean;
```
rmainseas, line 24: This could also probably do with a test case

src/Modals/Modal.tsx
```diff
@@ -75,6 +75,10 @@ export interface ModalProps {
    * The content to be displayed within the modal
    */
   children: ReactNode;
+  /**
+   * Identifies the element that provides the modal
+   */
+  ariaLabelledBy: string;
```
rmainseas, line 81: I think this is missing a test case

examples/Modals/Modal.md
```diff
@@ -20,6 +20,7 @@ const [modalVisible, setModalVisible] = useState(false);
     Open Modal
   </Button>
   <Modal
+    ariaLabelledBy="testButton"
```
rmainseas, line 23: There is no `testButton` in this example

examples/Modals/Modal.md
```diff
@@ -59,6 +60,7 @@ const [formValue, setFormValue] = useState('');
...
```
rmainseas, line 63: See above RE: `testButton`

src/Dropdown/Dropdown.tsx
rmainseas, line 18: I would extract this type out on it's own, or better yet - type this as [`HTMLOption[]`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement)

natalynyu: Good point! I added tests through 03a9141.

examples/Modals/Modal.md
natalynyu: Ah, my mistake. I updated this through a0c278a.

examples/Modals/Modal.md
natalynyu: I updated this as well through a0c278a.

src/Modals/Modal.tsx
natalynyu: I added a test to check the value of the aria-labelledBy property through commit 0a12491.

src/Dropdown/Dropdown.tsx
natalynyu: Good point - I added a test case for this through 1fa513b.

src/Dropdown/Dropdown.tsx
natalynyu: Ok good idea - I extracted this type out through 2f0f88b.

natalynyu: When I tried to type it as `HTMLOption`, it made all of the properties under HTMLOption mandatory, so I ended up extracting the type out instead

rmainseas: Ah, that's frustrating. Fair enough :)

src/Dropdown/__tests__/Dropdown.test.tsx
```diff
@@ -43,19 +53,51 @@ describe('Dropdown', function () {
   });
   it('contains the expected elements', function () {
     const dropdownOptionsCount = getAllByRole('option').length;
-    strictEqual(dropdownOptionsCount, 3);
+    strictEqual(dropdownOptionsCount, 4);
```
rmainseas, line 56: You might want to compare this with `options.length`, rather than a magic number :)

natalynyu: Ah! Good point. I updated this through commit c29d078.

PR #68, opened 2020-07-09
Modal and ModalHeader components - Trap Focus in Modal when Modal is Visible

natalynyu:
This pull request adds an optional `forwardRef` prop to the `Modal` and `ModalHeader`. The props specifies the ref of the modal, which is used to direct the focus. When the user clicks on the button to open the Modal, the focus should shift to the modal's header for accessibility purposes. This PR adds logic to redirect the focus to the first focusable element in the modal when the modal is visible, and it also traps the focus in the modal when the modal is visible.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#235](https://github.com/seas-computing/course-planner/issues/235)

### DIFF
--- a/examples/Modals/ModalHeader.md
+++ b/examples/Modals/ModalHeader.md
@@ -77,3 +77,85 @@ const [modalVisible, setModalVisible] = useState(false);
   </Modal>
 </>
 ```
+
+### With the `forwardRef`
+Ref example: The optional `forwardRef` property is set. When the test button is
+clicked, the focus shifts to the modal header.
+```jsx
+import { useState, useRef, } from 'react';
+import { VARIANT } from 'Theme';
+import {
+  Button,
+  Dropdown,
+  Modal,
+  ModalBody,
+} from 'mark-one';
+
+const RefExample = () => {
+  const ref = useRef(null);
+  const [value, setValue] = useState('');
+  const [modalVisible, setModalVisible] = useState(false);
+  const onButtonClick = () => {
+    setModalVisible(true);
+    /* Since modal may not have been rendered in DOM, wait for it to be
+    rendered by letting next task of event queue run first */
+    setTimeout(() => ref.current.focus(), 0);
+  }
+  return (
+    <>
+      <Button
+        onClick={onButtonClick}
+        variant={VARIANT.INFO}
+      >
+        Focus the input
+      </Button>
+      <Modal
+        ariaLabelledBy="testButton"
+        closeHandler={() => {setModalVisible(false)}}
+        isVisible={modalVisible}
+      >
+        <ModalHeader
+          closeButtonHandler={() => {setModalVisible(false)}}
+          tabIndex={0}
+          forwardRef={ref}
+        >
+          Modal Header
+        </ModalHeader>
+        <ModalBody>
+          <Dropdown
+            options={[
+              {
+                value: 'all',
+                label: 'All',
+              },
+              {
+                value: 'fall',
+                label: 'Fall',
+              },
+              {
+                value: 'spring',
+                label: 'Spring',
+              },
+            ]}
+            value={value}
+            id="semesters"
+            name="semesters"
+            onChange={function(event){
+              setValue(event.target.value);
+              alert('You changed the selection to ' + event.target.value);
+            }}
+            label="Semester"
+          />
+          <Button
+            onClick={() => setModalVisible(false)}
+            variant={VARIANT.BASE}
+          >
+            Close Modal
+          </Button>
+        </ModalBody>
+      </Modal>
+    </>
+  );
+};
+<RefExample />
+```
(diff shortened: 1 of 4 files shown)


src/Modals/Modal.tsx
```diff
@@ -100,6 +106,8 @@ export interface ModalProps {
    * e.g. initialize values in a form
    */
   onOpen?: () => void;
+  /** Specifies the ref of the Modal */
+  forwardRef?: RefObject<HTMLDivElement>;
```
farassadek, line 110: Reading the ref in react they did not emphasize that ref have to be to a div unless its our convention.

examples/Modals/ModalHeader.md
```diff
@@ -77,3 +77,85 @@ const [modalVisible, setModalVisible] = useState(false);
...
+
+const RefExample = () => {
+  const ref = useRef(null);
+  const [value, setValue] = useState('');
+  const [modalVisible, setModalVisible] = useState(false);
+  const onButtonClick = () => {
+    setModalVisible(true);
+    /* Since modal may not have been rendered in DOM, wait for it to be
+    rendered by letting next task of event queue run first */
+    setTimeout(() => ref.current.focus(), 0);
```
farassadek, line 102: for 0 timeout you can use ```setTimeout(() => ref.current.focus())```
Though I had bad experience with the setTimeout in general.

src/Modals/Modal.tsx
natalynyu: Jonathan advised in his other PR that 'Refs are generic so we need to provide a specific HTMLElement type to
use with it,' so I included `HTMLDivElement,` since for accessibility purposes the modal header (a div) should be focused for the screen reader.

examples/Modals/ModalHeader.md
natalynyu: Oh ok good to know. Yes I just needed a way to make sure the modal was rendered first.

natalynyu: I updated this through 8c72c64.

src/Modals/Modal.tsx
farassadek: make sense thanks

src/Modals/Modal.tsx
```diff
@@ -134,28 +142,49 @@ const Modal: FunctionComponent<ModalProps> = ({
...
       if (onOpen !== undefined) {
         onOpen();
       }
+      // Redirects the focus to the first focusable element in the modal when
+      // modal is visible
+      const listener = (event): void => {
+        const modal: HTMLElement = finalForwardRef.current;
+        if (!modal.contains(event.target as Node)) {
+          const focusables = 'button, [href], input, select, textarea,'
+            + ' [tabindex]:not([tabindex="-1"])';
```
farassadek, line 170: the focusable elements are buried inside the code so adding more elements would be a bit difficult.
It will be better if the list of the elements defined before the useEffect.

natalynyu: That makes sense. I moved the list of focusable elements out of the `useEffect` in commit 2914c42.

src/Modals/__tests__/Modal.test.tsx
```diff
@@ -50,6 +61,51 @@ describe('Modal', function () {
...
       });
     });
+    context('when forwardRef prop is present', function () {
+      beforeEach(function () {
+        const RefExample = () => {
+          const ref = useRef(null);
+          const [modalVisible, setModalVisible] = useState(false);
+          const onButtonClick = () => {
+            setModalVisible(true);
+            setTimeout(() => ref.current.focus(), 0);
```
farassadek, line 71: setTimeout(() => ref.current.focus());

natalynyu: Just fixed this through 4597089.

natalynyu: Thanks for catching this. Fixed this through 95e03a8!

natalynyu:
> Can you add a comment here about why are those the only focusable elements.
Good idea, I added this through 4c8109d.

src/Modals/__tests__/Modal.test.tsx
```diff
@@ -50,6 +61,60 @@ describe('Modal', function () {
...
     });
+    context('when forwardRef prop is present', function () {
+      beforeEach(function () {
+        const RefExample = () => {
+          const ref = useRef<HTMLDivElement>(null);
+          const [modalVisible, setModalVisible] = useState(false);
+          const onButtonClick = () => {
+            setModalVisible(true);
+            if (ref.current) {
+              ref.current.focus();
```
farassadek, line 72: The ref is always null therefore ref.current is always false and  ref.current.focus(); will never be executed.

PR #70, opened 2020-07-24
Fix absolute paths

jonseitz:
The absolute import for Theme doesn't work when the package is installed in an app. This has been happening when my editor auto-imports packages, so I'll need to fix that, or explore how we can add some kind of check for this.

Apologies for the inconvenience.


farassadek:
Looks good

farassadek:
How can I test this branch with the course-planner before I approve the changes ?

PR #71, opened 2020-07-27
Split sublevel theme types into new file

jonseitz:
The existing method of declaring all the theme types as part of the styled-components namespace was breaking when the library was published, since none of those themes were being exported and thus were missing when Theme/utils.ts tried to import them. This moves those sub-levels into the ThemeTypes.ts file (which will be compiled/exported), then declares the ValidThemeValues as part of the utils module itself.

As a quick and dirty way to test the compiled mark-one with course planner:

In local version of mark-one:
1. Remove any existing lib directory: `rm -rf ./lib`
2. compile the code: `npm run prepublishOnly`

In course-planner:
1. run `npm install` 
2. remove the existing lib directory from the installed version of mark-one: `rm -rf node_modules/mark-one/lib`
3. copy the built version of mark-one into node_modules: `cp -r /path/to/mark-one/lib node_modules/mark-one/lib`

Related to seas-computing/course-planner#242

### DIFF
--- a/src/Theme/ThemeTypes.ts
+++ b/src/Theme/ThemeTypes.ts
@@ -1,0 +1,58 @@
+export type ColorRange = {
+  light: string;
+  medium: string;
+  subtle?: string;
+  dark: string;
+  darker?: string;
+  accent?: string;
+};
+
+export type FontSpec = {
+  family?: string;
+  size?: string;
+  weight?: string;
+  color?: string;
+};
+
+export type AcademicArea = 'acs' |
+'am' |
+'ap' |
+'be' |
+'cs' |
+'ee' |
+'ese' |
+'general' |
+'mat & me' |
+'mde' |
+'msmba' |
+'sem';
+
+export type ColorCategory = 'background' | 'text' | 'area';
+
+export type FontCategory = 'base' |
+'body' |
+'data' |
+'note' |
+'bold' |
+'title'|
+'heading' |
+'error';
+
+export type WhiteSpaceSize = 'zero' |
+'xsmall' |
+'small' |
+'medium' |
+'large' |
+'xlarge';
+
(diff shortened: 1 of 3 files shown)


src/styled.d.ts
```diff
@@ -1,67 +1,17 @@
...
+import {
+  ColorRange,
+  TextColors,
+  AcademicArea,
+  FontCategory,
+  ShadowWeight,
+  BorderWeight,
+  WhiteSpaceSize,
+  FontSpec,
+} from './Theme/ThemeTypes';
```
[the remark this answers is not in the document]
jonseitz: `ColorCategory` isn't actually needed here, since we're just using the literal values of `background`, `text`, and `area` as keys in the object

src/Theme/ThemeTypes.ts
```diff
@@ -0,0 +1,58 @@
...
+'ap' |
+'be' |
+'cs' |
+'ee' |
+'ese' |
+'general' |
+'mat & me' |
+'mde' |
+'msmba' |
+'sem';
```
farassadek, line 28: AcademicArea is specific to the course planner, not sure if they should be in mark-one.

jonseitz: Yeah, I've thought this too. Originally the idea was that we'd have the individual area colors as part of Mark One so we can use them across apps, since they're the same colors used on the SEAS website. But it does add the complication of us needed to add any new colors in Mark One before they can be used anywhere else. 

So it's a fair point, but I don't think it should be addressed as part of this ticket since we're just trying to fix a bug a restore functionality. Can you open a new issue on the Course Planner project to address this?

farassadek: Agree, I will open a new ticket issue for this one.

PR #74, opened 2020-08-25
Solve for Modal Content Overflow

natalynyu:
This PR adds the CSS property `overflow` to the `ModalBody` so that if the modal contents are larger than the modal itself, the user will be able to scroll through the contents. I added the CSS property to the body instead of the overall modal component so that the title of the modal would be 'sticky' and could always be seen.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#235](https://github.com/seas-computing/course-planner/issues/235)

### DIFF
--- a/src/Modals/ModalBody.tsx
+++ b/src/Modals/ModalBody.tsx
@@ -10,7 +10,7 @@ interface ModalBodyProps {
 
 const StyledModalBody = styled.div<ModalBodyProps>`
   padding: ${({ theme }): string => theme.ws.medium};
-  overflow: scroll;
+  overflow: auto;
 `;
 
 /**
--- a/src/Modals/ModalBody.tsx
+++ b/src/Modals/ModalBody.tsx
@@ -10,6 +10,7 @@ interface ModalBodyProps {
 
 const StyledModalBody = styled.div<ModalBodyProps>`
   padding: ${({ theme }): string => theme.ws.medium};
+  overflow: scroll;
 `;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Modals/ModalBody.tsx
```diff
@@ -10,6 +10,7 @@ interface ModalBodyProps {
 
 const StyledModalBody = styled.div<ModalBodyProps>`
   padding: ${({ theme }): string => theme.ws.medium};
+  overflow: scroll;
```
rmainseas, line 13: `overflow: auto` might be better to hide the scrollbar unless needed

natalynyu: Interesting, does the scrollbar always show up for you? It only shows up for me when there's an overflow!

natalynyu: I updated the overflow property to `auto` through commit 2437447.

rmainseas: Yeah, I suspect that's MacOS being fancy :)

Generally it shows up all the time if you use `overflow: scroll`, whereas `overflow: auto` has the behavior you're describing.

https://stackoverflow.com/questions/6689412/difference-between-html-overflow-auto-and-overflow-scroll#:~:text=Auto%20will%20only%20show%20a,and%20you%20cant%20scroll%20it.&text=overflow%3A%20scroll%20will%20hide%20all,on%20the%20element%20in%20question.

rmainseas:
Approved.

src/Modals/ModalBody.tsx
natalynyu: Ah good to know thanks for sharing that!

PR #75, opened 2020-10-13
Update formatting for RadioButton and Checkbox components

natalynyu:
The purpose of this PR is to fix the formatting for the `RadioButton` and `Checkbox` components so that when they are used in a form in conjunction with the `TextInput` component, they are all appropriately aligned at the same start point. With the previous state of these components, this is how the components look when used together:

image

I added `TextInput` and `RadioButton` components to the `ModalBody` overflow example to demonstrate the alignment of the components with the updated code:
image

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#229](https://github.com/seas-computing/course-planner/issues/229)

### DIFF
--- a/examples/Forms/CheckLabel.md
+++ b/examples/Forms/CheckLabel.md
@@ -4,118 +4,116 @@ Aligning `RadioButton` and `Dropdown` fields example: The labelPosition for `Rad
 ```jsx
 import { useState } from 'react';
 import { Fieldset, RadioButton, Dropdown, POSITION } from 'mark-one';
-const Example = () => {
-  const [checkedValue, setCheckedValue] = useState('');
-  const [value, setValue] = useState('fall');
-  return (<>
-    <Fieldset
-      legend="Fieldset Legend"
-      isBorderVisible={false}
-      isLegendVisible={false}
-    >
-      <RadioButton
-        label="Foo"
-        value="foo"
-        name="nonVisibleFieldsetTest"
-        checked={checkedValue === 'foo'}
-        onChange={(event) => setCheckedValue(event.target.value)}
-        labelPosition={POSITION.RIGHT}
-      />
-      <RadioButton
-        label="Bar"
-        value="bar"
-        name="nonVisibleFieldsetTest"
-        checked={checkedValue === 'bar'}
-        onChange={(event) => setCheckedValue(event.target.value)}
-        labelPosition={POSITION.RIGHT}
-      />
-      <Dropdown
-        options={[
-          {
-            value: 'all',
-            label: 'All',
-          },
-          {
-            value: 'fall',
-            label: 'Fall',
-          },
-          {
-            value: 'spring',
-            label: 'Spring',
-          },
-        ]}
-        value={value}
-        id="semesters"
-        name="semesters"
-        onChange={function(event){
-          setValue(event.target.value);
-          alert('You changed the selection to ' + event.target.value);
-        }}
-        label="Semester"
-        labelPosition={POSITION.TOP}
-      />
-    </Fieldset>
-  </>);
-}
-<Example />
+
+const [checkedValue, setCheckedValue] = useState('');
+const [value, setValue] = useState('fall');
+
+<>
+  <Fieldset
+    legend="Fieldset Legend"
+    isBorderVisible={false}
+    isLegendVisible={false}
+  >
+    <RadioButton
+      label="Foo"
+      value="foo"
+      name="nonVisibleFieldsetTest"
+      checked={checkedValue === 'foo'}
+      onChange={(event) => setCheckedValue(event.target.value)}
+      labelPosition={POSITION.RIGHT}
+    />
+    <RadioButton
+      label="Bar"
+      value="bar"
+      name="nonVisibleFieldsetTest"
+      checked={checkedValue === 'bar'}
+      onChange={(event) => setCheckedValue(event.target.value)}
+      labelPosition={POSITION.RIGHT}
+    />
+    <Dropdown
+      options={[
+        {
+          value: 'all',
+          label: 'All',
+        },
+        {
+          value: 'fall',
+          label: 'Fall',
+        },
+        {
+          value: 'spring',
+          label: 'Spring',
+        },
+      ]}
+      value={value}
+      id="semesters"
+      name="semesters"
+      onChange={function(event){
+        setValue(event.target.value);
+        alert('You changed the selection to ' + event.target.value);
+      }}
+      label="Semester"
+      labelPosition={POSITION.TOP}
+    />
+  </Fieldset>
+</>
 ```
 
 Aligning `RadioButton` and `Dropdown` fields example: The labelPosition for `RadioButton` and `Dropdown` is `POSITION.LEFT`.
 ```jsx
 import { useState } from 'react';
 import { Fieldset, RadioButton, Dropdown, POSITION } from 'mark-one';
-const Example = () => {
-  const [checkedValue, setCheckedValue] = useState('');
-  const [value, setValue] = useState('fall');
-  return (<>
-    <Fieldset
-      legend="Fieldset Legend"
-      isBorderVisible={false}
-      isLegendVisible={false}
-    >
-      <RadioButton
-        label="Foo"
-        value="foo"
-        name="nonVisibleFieldsetTest"
-        checked={checkedValue === 'foo'}
-        onChange={(event) => setCheckedValue(event.target.value)}
-        labelPosition={POSITION.LEFT}
-      />
-      <RadioButton
-        label="Bar"
-        value="bar"
-        name="nonVisibleFieldsetTest"
-        checked={checkedValue === 'bar'}
-        onChange={(event) => setCheckedValue(event.target.value)}
-        labelPosition={POSITION.LEFT}
-      />
-      <Dropdown
-        options={[
-          {
-            value: 'all',
-            label: 'All',
-          },
-          {
-            value: 'fall',
-            label: 'Fall',
-          },
-          {
-            value: 'spring',
-            label: 'Spring',
-          },
-        ]}
-        value={value}
-        id="semesters"
-        name="semesters"
-        onChange={function(event){
-          setValue(event.target.value);
-          alert('You changed the selection to ' + event.target.value);
-        }}
-        label="Semester"
-        labelPosition={POSITION.LEFT}
-      />
-    </Fieldset>
-  </>);
-}
-<Example />
+
+const [checkedValue, setCheckedValue] = useState('');
+const [value, setValue] = useState('fall');
+
+<>
+  <Fieldset
+    legend="Fieldset Legend"
+    isBorderVisible={false}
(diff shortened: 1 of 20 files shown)


jonseitz (changes requested):
Unfortunately changing the `Label` component like this will cause some problems with our other inputs, since the label size will shrink/grow to fit the text. The goal was to have all of the input boxes lined up, regardless of the label size. 
 


This is kind of similar to the issue @farassadek is having with the filter inputs, where our `Label` component has been set up for the specific use case of text/select inputs in forms. As in that case, I think we need to either provide a parameter that will control the display of the label for these non-standard cases, or create a separate label component that we can use for check/radio inputs.

Personally, I think we're approaching a level of complexity where it would make sense to split some of these use cases out into their own components, but I'm happy to discuss further.

natalynyu:
@jonseitz Thanks for your feedback! Those are good points. I was wondering: in the new implementation of the `Label` component, should the input fields still line up (and if so, would the text of the longer descriptions below wrap to the next line)?  
image

jonseitz:
@natalynyu Yes, the current behavior is to wrap the line:



Just to clarify though, my suggestion for a new implementation of the `Label` was that we'd be creating one that would **only** be used for checkboxes/radio buttons, so I wasn't thinking that the new `Label` would affect the text inputs at all. Does that make sense?

natalynyu:
@jonseitz Ah right! I confused myself - sorry about that. I understand now - so we'd be keeping the original behavior of the label in relation to the `Text Input` component.

natalynyu:
@jonseitz For the example screenshot that you posted, where would the radio/checkboxes be located in relation to the group of text inputs? 
image
Would the radio and checkbox labels in these cases line up with the text input field?
image
image


For the case of the course admin modal, though, the text label position is `POSITION.TOP` as opposed to `POSITION.LEFT` in the example above. I think for the modal we would want each line to be left-aligned: 
image

jonseitz:
Hmmmm, very good questions. Thinking just about the checkboxes/radios, my sense is:

- Whenever `labelPosition={POSITION.RIGHT}`, the checkbox and label should be adjacent and flush all the way left
- When `labelPosition={POSITION.LEFT}` the grid should be the same as the text inputs, with label in line with the other labels and the checkbox in line with the inputs
- When `labelPosition={POSITION.TOP}` both should be flush left

So maybe it's really just the case of `POSITION.RIGHT` where we need to make changes in the `Label`?

jonseitz:
I think we should handle the check/radio and input/select separately -- It's just going to be too complicated to try to make every possible combination of labels/inputs line up correctly.

So really, I think we just need to change the `POSITION.RIGHT` option for the check/radio so that it's flush left, and then we should add a note to our docs that we should use that versions whenever we're using it in conjunction with a select/input that uses `POSITION.TOP`. What do you think about that?

natalynyu:
@jonseitz Ok, yes I think accounting for all situations at once may be too complicated. I'm not really sure what you meant by your suggestion. Do you mean making a second `Label` component that is just used for the case when the text input has a labelPosition of `POSITION.TOP` & the check/radio's label is `POSITION.RIGHT`/`POSITION.LEFT`?

jonseitz:
Right now, we have four components that are all using the same `Label`

- `TextInput`
- `Dropdown`
- `Checkbox`
- `RadioButton`

I'm proposing that we:

1. Make no changes to the `TextInput` and `Dropdown` components
2. Change the name of `Label` (e.g. `InputLabel`), allowing `POSITION.TOP` and `POSITION.LEFT`, and only use that in `TextInput` and `Dropdown`
3. Create a new component, e.g. `CheckLabel`, allowing `POSITION.LEFT` and `POSITION.RIGHT`, and use that in place of `Label` in the the `Checkbox` and `RadioButton` components.

In the `InputLabel`: `POSITION.TOP` and `POSITION.LEFT` would stay exactly as they are

In the `CheckLabel`: we'd switch to using a flush-left style, like what you'd originally done here, with `POSITION.RIGHT` and leave `POSITION.LEFT` as it is.

That way, if you're using a `TextInput` with `POSITION.TOP`, you can use a `Checkbox` with `POSITION.RIGHT` beneath it, and both input "boxes" will be flush-left.

Then if you have a `TextInput` with `POSITION.LEFT`, you can use a `Checkbox` with `POSITION.LEFT` beneath it, and the labels and inputs will line up.

---

Technically, this doesn't actually require us to create separate label components. However, I feel like we've already added a lot of conditional logic to make that a single `Label` adapt to all the different possible `POSITION` values with the two types of inputs, and I'm concerned about longterm maintainability as we add more components down the line. Ultimately we're not using the `Label` elements directly when we import `mark-one` into the app, so this shouldn't affect the public API of the library at all.

natalynyu:
@jonseitz thanks for your help on this! I made the `CheckLabel` component and renamed the original `Label` to `InputLabel.` I also added docs a few more tests, and I exported the `POSITION` enum from `InputLabel` so that it can be used in the docs.

examples/Forms/CheckLabel.md
```diff
@@ -0,0 +1,121 @@
+The `CheckLabel` component is used by the `RadioButton` and `Checkbox` components, while the `InputLabel` component is used by the `TextInput` and `Dropdown` components. These label variations were created to align input fields when using various form elements in tandem.
+
+Aligning `RadioButton` and `Dropdown` fields example: The labelPosition for `RadioButton` is `POSITION.RIGHT`, and the `labelPosition` for the `Dropdown` is `POSITION.TOP`.
+```jsx
+import { useState } from 'react';
+import { Fieldset, RadioButton, Dropdown, POSITION } from 'mark-one';
+const Example = () => {
```
jonseitz, line 7: For all of these examples, you can call `useState` directly -- you don't need to wrap everything in the `<Example>` component. It's not necessary wrong, but we should try to keep our example code as clean as possible.

examples/Modals/ModalBody.md
```diff
@@ -56,79 +65,76 @@ const [modalVisible, setModalVisible] = useState(false);
...
-          disabled
-          checked
+          <TextInput
+            value="test"
+            name="example"
+            id="example"
+            label="Description:"
+            disabled="disabled"
+            onChange={() => {}}
           />
```
jonseitz, line 124: This example doesn't match up with our `CheckLabel` and `InputLabel` examples. I think we'd want this `TextInput` to be `position={POSITION.TOP}`

src/Forms/CheckLabel.tsx
```diff
@@ -0,0 +1,134 @@
...
+  disabled?: boolean;
+}
+
+export interface LabelProps {
+  /** The id of the field tied to this label */
+  htmlFor: string;
+  /** Specifies the label text */
+  label: string;
+  /** Allows you to pass in a label position property from the POSITION enum */
+  labelPosition?: POSITION;
```
jonseitz, line 34: For the checkboxes, I don't think we really need the `POSITION.TOP` option, so it would make sense to limit this to type `POSITION.LEFT | POSITION.RIGHT`. This would probably also need to be done in the `Checkbox` and `RadioButton` components.

src/Forms/CheckLabel.tsx
```diff
@@ -0,0 +1,134 @@
...
+  margin: ${fromTheme('ws', 'small')};
+  align-items: baseline;
+  gap: ${({ theme }) => (theme.ws.xsmall) + ' ' + (theme.ws.xsmall)};
+`;
+
+const StyledCheckLabelText = styled.span<StyledCheckLabelTextProps>`
+  display: ${({ isLabelVisible }) => (isLabelVisible ? 'inline' : 'none')};
+  grid-area: label;
+  justify-self: ${({ labelPosition }) => (
+    (labelPosition === POSITION.TOP || labelPosition === POSITION.RIGHT)
```
jonseitz, line 79: As noted above, we can probably leave `POSITION.TOP` out of the Checkboxes

src/Forms/CheckLabel.tsx
```diff
@@ -0,0 +1,134 @@
...
+    return `"input input input"
+            "e e e"`;
+  }
+  if (labelPosition === POSITION.RIGHT) {
+    return `"input label label"
+            ". e e"`;
+  }
+  return `"label input input"
+          ". e e"`;
+};
```
jonseitz, line 57: We should use consistent `grid-template-area` ids between this and `InputLabel`, so they should either both be `i`, `l`, and `e`, or both be `input`, `label`, and `error`.

src/Forms/InputLabel.tsx
```diff
@@ -31,7 +31,7 @@ export interface StyledLabelTextProps {
   disabled?: boolean;
 }
 
-export interface LabelProps {
+export interface InputLabelProps {
```
jonseitz, line 34: Not part of the diff, but similar to the `CheckLabel` we should probably limit the allowed values of labelPosition to `POSITION.LEFT | POSITION.TOP` here and in the `TextInput` and `Dropdown` components

jonseitz (changes requested):
Just a couple of general notes. On the examples I didn't flag every instance of this, but we typically don't need to wrap our code in the `Example` components, even if we're using hooks. We should be able to write everything in the code block and it should render fine.

I also noted that we should define types for the `labelPosition` props that only allows the expected values for each of the labels (i.e `TOP` and `LEFT` for InputLabel, `LEFT` and `RIGHT` for `CheckLabel`) That's going to need to change in a couple places, I expect, but not everything was included in the PR so I couldn't make notes in the gh interface.

src/Forms/CheckLabel.tsx
natalynyu: Ah my bad, originally I was thinking that `grid-area` would control which label (`InputLabel` vs. `Checklabel`) got used. I updated this through b9fd9aa to use `l` and `i` across components instead.

src/Forms/CheckLabel.tsx
natalynyu: Good point - I added a type called `CheckLabelPosition` so I didn't have to keep putting `POSITION.LEFT | POSITION.RIGHT` in multiple places. (commit 673b37e)

src/Forms/CheckLabel.tsx
natalynyu: This was also updated in 673b37e (above).

src/Forms/InputLabel.tsx
natalynyu: Agreed, I added the type `InputLabelPosition` in commit c4c61a3.

examples/Modals/ModalBody.md
natalynyu: D'oh. I updated this in 5792c23.

examples/Forms/CheckLabel.md
natalynyu: That's a great suggestion! I found that a few other Form element examples were also wrapping in an Example component, so I updated those as well through 1f1c998. Definitely a lot cleaner!!

jonseitz (approved):
This looks good!

PR #76, opened 2020-10-15
Create ExternalLink component

jonseitz:
Our existing Link component is a wrapper around ReactRouter's Link, so we can only use it for navigation within the app. This ExternalLink is just a `styled.a` with CSS to match the Link, so that we can create links to external pages that look similar.

The specific use case for this is the "log out" link from #246

### DIFF
--- a/examples/Links/ExternalLink.md
+++ b/examples/Links/ExternalLink.md
@@ -1,0 +1,12 @@
+This component is a styled version of a regular `<a>` link, for use when we need to link to external resources.
+
+```jsx
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSignOutAlt } from '@fortawesome/free-solid-svg-icons';
+
+<ExternalLink href="https://key.harvard.edu/logout">
+  Log Out
+  {' '}
+  <FontAwesomeIcon icon={faSignOutAlt} />
+</ExternalLink>
+```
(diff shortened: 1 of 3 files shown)


natalynyu (approved):
This looks good to me! I just had one minor comment about the use of the icon in conjunction with the text in the link. I was also thinking of whether external links would ever need to be a different color, but I don't think so.

examples/Links/ExternalLink.md
```diff
@@ -0,0 +1,12 @@
+This component is a styled version of a regular `<a>` link, for use when we need to link to external resources.
+
+```jsx
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSignOutAlt } from '@fortawesome/free-solid-svg-icons';
+
+<ExternalLink href="https://key.harvard.edu/logout">
+  Log Out
+  {' '}
```
natalynyu, line 9: Originally I was thinking that maybe the space could be integrated into the component definition, but this seems to be a rare enough case to be using both text and an icon. Also, I noticed that when you hover, the underline extends a little past the text due to the space, but this is a very tiny detail - 
Screen Shot 2020-10-15 at 2 51 51 PM

jonseitz: That is weird. I suspect there's something in the FontAwesome library that's removing the underline. In any case, I'd prefer to leave handling of the link content to the developer, otherwise we'll drive ourselves crazy trying to handle edge cases.

Re: link color, I believe we had discussed this with Vittorio and decided to use a consistent color for all of links, so I agree that we wouldn't want to provide other color options.

PR #77, opened 2020-10-15
Feature/224 add hide error prop to input text label

farassadek:
New Feature: 
Adding a hideError Prop to the the TextInput.

hideError will be used by the TextInput to hide the error if exist.
It will be used by the label to adjust the style of the area of the errorMessage(will remove the spaces under the TextInput).

Tests of the hideError is included in the InputText test.

### DIFF
--- a/examples/Forms/TextInput.md
+++ b/examples/Forms/TextInput.md
@@ -72,7 +72,6 @@ const [value, setValue] = useState('Some valid input');
   }}
 />
 ```
-
 Text Input example: The `disabled` prop set to `true.`
 
 ```jsx
@@ -104,6 +103,48 @@ const [value, setValue] = useState('Some valid input');
   name="example"
   id="example"
   label="Description:"
+  onChange={(event) => {
+    setValue(event.target.value);
+  }}
+/>
+```
+
+Label example: The optional `errorMessage` prop is not set, and the 
+`hideError` prop is set to `true.` Remove the `errorMesage` style.
+
+```jsx
+import { useState } from 'react';
+
+const [value, setValue] = useState('Some valid input');
+
+<TextInput
+  value={value}
+  name="example"
+  id="example"
+  label="Description:"
+  hideError={true}
+  onChange={(event) => {
+    setValue(event.target.value);
+  }}
+/>
+```
+
+Label example: The optional `errorMessage` prop is set, and the 
+`hideError` prop is set to `true.` Remove the `errorMesage` and
+its style.
+
+```jsx
+import { useState } from 'react';
+
+const [value, setValue] = useState('Some valid input');
+
+<TextInput
+  value={value}
+  name="example"
+  id="example"
+  label="Description:"
+  errorMessage="Error: Please enter a valid ID"
+  hideError={true}
   onChange={(event) => {
     setValue(event.target.value);
   }}
--- a/examples/Forms/TextInput.md
+++ b/examples/Forms/TextInput.md
@@ -110,7 +110,7 @@ const [value, setValue] = useState('Some valid input');
 ```
 
 Label example: The optional `errorMessage` prop is not set, and the 
-`hideError` prop is set to `true.` Remove the `errorMesage` style.
+`hideError` prop is set to `true`, which removes the errorMessage style.
 
 ```jsx
 import { useState } from 'react';
@@ -130,8 +130,8 @@ const [value, setValue] = useState('Some valid input');
 ```
 
 Label example: The optional `errorMessage` prop is set, and the 
-`hideError` prop is set to `true.` Remove the `errorMesage` and
-its style.
+`hideError` prop is set to `true`, which removes the errorMessage 
+and its style.
 
 ```jsx
 import { useState } from 'react';
--- a/examples/Forms/TextInput.md
+++ b/examples/Forms/TextInput.md
@@ -74,6 +74,7 @@ const [value, setValue] = useState('Some valid input');
   }}
 />
 ```
+
 Text Input example: The `disabled` prop set to `true.`
 
 ```jsx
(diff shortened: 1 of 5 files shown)


examples/Forms/TextInput.md
```diff
@@ -110,6 +109,48 @@ const [value, setValue] = useState('Some valid input');
 />
 ```
 
+Label example: The optional `errorMessage` prop is not set, and the 
+`hideError` prop is set to `true.` Remove the `errorMesage` style.
```
natalynyu, line 113: Typo: `errorMessage`
I would also change this to "`hideError` prop is set to `true,` which removes the `errorMessage` style."

examples/Forms/TextInput.md
```diff
@@ -110,6 +109,48 @@ const [value, setValue] = useState('Some valid input');
...
+  hideError={true}
+  onChange={(event) => {
+    setValue(event.target.value);
+  }}
+/>
+```
+
+Label example: The optional `errorMessage` prop is set, and the 
+`hideError` prop is set to `true.` Remove the `errorMesage` and
+its style.
```
natalynyu, line 134: Typo: `errorMessage`
I would also change this to "`hideError` prop is set to `true,` which removes the `errorMessage` and its style."

src/Forms/TextInput.tsx
```diff
@@ -39,6 +39,8 @@ export interface TextInputProps {
   isRequired?: boolean;
   /** Specifies the ref of the text input */
   forwardRef?: Ref<HTMLInputElement>;
+  /** If true, ommit the margin style in label */
```
natalynyu, line 42: Typo: omit

src/Forms/Label.tsx
```diff
@@ -74,9 +80,16 @@ const StyledLabel = styled.label<StyledLabelProps>`
   grid-template-areas: ${({ labelPosition, isLabelVisible }) => (
     generateGrid(labelPosition, isLabelVisible)
   )};
-  margin: ${fromTheme('ws', 'small')};
+  margin: ${fromTheme('ws', 'small')}
   align-items: baseline;
   gap: ${({ theme }) => (theme.ws.xsmall) + ' ' + (theme.ws.xsmall)};
+  ${({ hideError }) => (hideError
+    ? `margin: 0px;
```
natalynyu, line 87: I think that we would still want to have the `margin` for the label in the case that `hideError` is true. 
image

src/Forms/Label.tsx
```diff
@@ -74,9 +80,16 @@ const StyledLabel = styled.label<StyledLabelProps>`
...
+       gap: 0px;
+       grid-template-rows: 1fr;
+      `
+    : null
```
natalynyu, line 91: Since this `null` value is in a tagged template literal, it would be converted to the string "null."

image

I think better here instead of `null` would be just an empty string "".

natalynyu (changes requested):
Looks good! I just had a few comments.

examples/Forms/TextInput.md
farassadek: What is the type?

src/Forms/Label.tsx
farassadek: OK will take care of this

examples/Forms/TextInput.md
natalynyu: It says "errorMesage" instead of "errorMessage"

src/Forms/__tests__/TextInput.test.tsx
```diff
@@ -146,13 +184,9 @@ describe('Text input', function () {
...
       strictEqual(defaultValue, 'Spring');
     });
-    it('does not render the error message', function () {
-      const errorField = queryByText('error');
-      strictEqual(errorField, null);
-    });
-    it('does not set aria-invalid', function () {
-      const input = getByRole('textbox');
-      strictEqual(input.hasAttribute('aria-invalid'), false);
+    it('gap and margin and grid-template-rows style', function () {
```
natalynyu, line 187: This is a small point, but the test description should start with a verb so that the test description is like a sentence. The context says "When hideError prop is not set," and then you would complete the sentence with your test description, e.g. 'should set the gap, margin, and grid-template-rows styles accordingly' or something like that.

src/Forms/Label.tsx
```diff
@@ -74,9 +80,16 @@ const StyledLabel = styled.label<StyledLabelProps>`
   grid-template-areas: ${({ labelPosition, isLabelVisible }) => (
     generateGrid(labelPosition, isLabelVisible)
   )};
-  margin: ${fromTheme('ws', 'small')};
+  margin: ${fromTheme('ws', 'small')}
```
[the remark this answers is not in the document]
farassadek: Fixed by pushing the update that was not pushed before

examples/Forms/TextInput.md
farassadek: oh i see

src/Forms/Label.tsx
```diff
@@ -77,6 +83,13 @@ const StyledLabel = styled.label<StyledLabelProps>`
   margin: ${fromTheme('ws', 'small')};
   align-items: baseline;
   gap: ${({ theme }) => (theme.ws.xsmall) + ' ' + (theme.ws.xsmall)};
+  ${({ hideError }) => (hideError
```
jonseitz, line 86: I know I'm not the reviewer on this anymore, but one concern I had when I initially looked at this PR is that our style so far has been to apply conditionals like this on individual rules (as in `margin: ${hideError ? 0px: 10px};`) rather than adding them to the bottom and relying on the cascade to override the rules. 

The second method does work, but it makes it harder to maintain these components if we need to add more conditional rendering in the future. I'll always argue for consistency, so I would prefer if we could change this to use the conditional on individual lines instead.

natalynyu: That's a good point I didn't think of.

farassadek: This is a good point though it will increase the complexity if we add rules and condition per a property.

src/Forms/__tests__/TextInput.test.tsx
```diff
@@ -111,6 +111,124 @@ describe('Text input', function () {
...
+    });
+    it('renders the correct default value', function () {
+      const inputField = document.getElementById('semester') as HTMLInputElement;
+      const defaultValue = inputField.value;
+      strictEqual(defaultValue, 'Spring');
+    });
+    it('does not render the error message', function () {
+      const errorField = queryByText('error');
+      strictEqual(errorField, null);
+    });
```
natalynyu, line 149: Is this the behavior we want that when there is an error message we don't show it? This seems like it may be an issue to hide the error when there is one. I thought the purpose of the `hideError` prop was to  adjust the formatting when there is no `errorMessage` present so that there isn't an empty space under the filter fields.

natalynyu (changes requested):
This looks good - Mostly, I'm just looking for clarification on whether the error message should be hidden if `errorMessage` is defined and `hideError` is true. I'm not sure if this is the behavior we want, since I think we would want to show the user an error message if there was one?

src/Forms/__tests__/TextInput.test.tsx
farassadek: We did not discuss that, therefore I (logic of not native english speaker) decide to hide the error message when hideError is set to true. Otherwise "hideError" will be misleading if we hide a non existing error.

src/Forms/__tests__/TextInput.test.tsx
```diff
@@ -111,6 +111,124 @@ describe('Text input', function () {
...
+    it('set grid-template-rows styles accordingly', function () {
+      const style = window.getComputedStyle(getByText('semester').parentNode as HTMLElement);
+      strictEqual(style['grid-template-rows'], '1fr');
+    });
+  });
+  context('when hideError is set and isLabelVisible is not ', function () {
```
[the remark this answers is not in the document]
farassadek: hideError and isLabelVisible controls the gap.
If hideError = true and isLabelVisible = false then gap should be 0.

src/Forms/__tests__/TextInput.test.tsx
natalynyu: J Seitz confirmed that we should be able to hide the error when `errorMessage` is true.

natalynyu (approved):
Looks good to me

PR #79, opened 2020-11-05
adding aria-describedby to inputText and Dropdown component

farassadek:
Adding `aria-describedby` to InputText and Dropdown for accessibility. 
Linting 
No testing required

### DIFF
--- a/src/Forms/Dropdown.tsx
+++ b/src/Forms/Dropdown.tsx
@@ -90,6 +90,7 @@ const Dropdown: FunctionComponent<DropdownProps> = (props): ReactElement => {
         aria-errormessage={`${id}-error`}
         aria-invalid={errorMessage ? true : null}
         aria-required={isRequired}
+        aria-describedby={name}
         ref={forwardRef}
       >
         {options.map((option) => (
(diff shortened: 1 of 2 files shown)


jonseitz (changes requested):
Looking at [the spec for `aria-describedby`](https://www.w3.org/TR/WCAG20-TECHS/ARIA1.html), I don't think it's being implemented correctly here. The value is supposed to be the `id` field of a separate element that provides more descriptive information about the field, like any validation rules, etc. So unless we know that we'll always have an element on the page whose `id` is the same as the value of `name`, this isn't going to work correctly. Or that's my interpretation anyway -- @Vittorio2015 may be able to provide some more info here.

What is the intended goal of adding `aria-describedby` here? If it's just to provide an accessible label I think `aria-label` would be more appropriate. But since we do already have the `InputLabel` component associated with the field and that should provide an accessible label already.

Vittorio2015:
Moving the information about the functionality of the filter works well when placed as a `label` (and removing both 'aria-describedby' and the 'span' with the description, as shown in this screen capture), although we should add to this specific label a reference to the course title, such as "The table will be filtered by course title, as characters are typed in this input form"
Here is what I tested with Mac OS screen reader VoiceOver ([movie available on Slack](https://seasarc.slack.com/files/UJKHHHJTB/F01E4E71E3X/courseplanner_input_label.mov?origin_team=T02FTFW9V&origin_channel=CL2EHB5S4)):
image

PR #80, opened 2020-11-13
Add Mark One Props Interface

natalynyu:
This interface was added as a result of feedback from the course planner [absence modal PR ](https://github.com/seas-computing/course-planner/pull/276). In that ticket, I was using `document.getElementById` to get a reference to button elements to focus them. Jonathan Seitz recommended that we should be using `ref` instead. We had a discussion on this and concluded that since the properties `ref` and `aria-label` will eventually be used across many different `mark one` components, it would make sense to make an interface for these components at the top level of `mark one` so that component props could extend the interface to include these common props. We decided it made sense to only implement the interface for the components needed in the absence modal for now, which would be the button components. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#252](https://github.com/seas-computing/course-planner/issues/252)

### DIFF
--- a/examples/Buttons/BorderlessButton.md
+++ b/examples/Buttons/BorderlessButton.md
@@ -86,7 +86,10 @@ import { faTrash } from '@fortawesome/free-solid-svg-icons';
 ### With the `forwardRef`
 Ref example: The optional `forwardRef` property is set. When the primary themed button is clicked, the focus shifts to the borderless button.
 ```jsx
-import { useState, useRef, } from 'react';
+import {
+  useState,
+  useRef,
+} from 'react';
 import {
   Button,
   VARIANT,
--- a/examples/Buttons/BorderlessButton.md
+++ b/examples/Buttons/BorderlessButton.md
@@ -118,6 +118,7 @@ const BorderlessButtonRefExample = () => {
       variant={VARIANT.DANGER}
       tabIndex={0}
       forwardRef={ref}
+      alt="testing"
     >
       <FontAwesomeIcon icon={faTrash} size="lg" />
     </BorderlessButton>
--- a/examples/Buttons/BorderlessButton.md
+++ b/examples/Buttons/BorderlessButton.md
@@ -118,7 +118,7 @@ const BorderlessButtonRefExample = () => {
       variant={VARIANT.DANGER}
       tabIndex={0}
       forwardRef={ref}
-      alt="testing"
+      alt="testing delete"
     >
       <FontAwesomeIcon icon={faTrash} size="lg" />
     </BorderlessButton>
--- a/examples/Buttons/BorderlessButton.md
+++ b/examples/Buttons/BorderlessButton.md
@@ -101,7 +101,7 @@ const BorderlessButtonRefExample = () => {
   const ref = useRef(null);
   const [value, setValue] = useState('');
   const onButtonClick = () => {
-    setTimeout(() => ref.current.focus());
+    ref.current.focus();
   }
   return (
     <>
--- a/examples/Buttons/BorderlessButton.md
+++ b/examples/Buttons/BorderlessButton.md
@@ -11,7 +11,7 @@ import { faArrowUp } from '@fortawesome/free-solid-svg-icons';
   }}
   variant={VARIANT.BASE}
 >
-  <FontAwesomeIcon icon={faArrowUp} size="small" />
+  <FontAwesomeIcon icon={faArrowUp} size="lg" />
 </BorderlessButton>
 ```
 
@@ -28,7 +28,7 @@ import { faEnvelope } from '@fortawesome/free-solid-svg-icons';
   }}
   variant={VARIANT.PRIMARY}
 >
-  <FontAwesomeIcon icon={faEnvelope} size="small" />
+  <FontAwesomeIcon icon={faEnvelope} size="lg" />
 </BorderlessButton>
 ```
 
@@ -45,7 +45,7 @@ import { faTimes } from '@fortawesome/free-solid-svg-icons';
   }}
   variant={VARIANT.SECONDARY}
 >
-  <FontAwesomeIcon icon={faTimes} size="small" />
+  <FontAwesomeIcon icon={faTimes} size="lg" />
 </BorderlessButton>
 ```
 
@@ -62,7 +62,7 @@ import { faQuestion } from '@fortawesome/free-solid-svg-icons';
   }}
   variant={VARIANT.INFO}
 >
-  <FontAwesomeIcon icon={faQuestion} size="small" />
+  <FontAwesomeIcon icon={faQuestion} size="lg" />
 </BorderlessButton>
 ```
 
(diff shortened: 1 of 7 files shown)


farassadek:
Based on React docs (https://reactjs.org/docs/forwarding-refs.html) the on the Note for component library maintainers section `When you start using forwardRef in a component library, you should treat it as a breaking change and release a new major version of your library.`
Therefore I think should be a braking change.

src/Theme/MarkOneWrapper.tsx
```diff
@@ -1,10 +1,26 @@
...
 
+/**
+ * This interface includes props that are common to many components for testing
+ * and accessibility purposes.
+ */
+export interface MarkOneProps<T extends HTMLElement> {
+  /** Specifies the ref of the element */
+  forwardRef?: Ref<T>;
+  /** Corresponds to HTML attribute tabindex */
+  tabIndex?: number;
```
farassadek, line 19: I did not understand the usage of the tabIndex here

examples/Buttons/BorderlessButton.md
```diff
@@ -79,6 +79,51 @@ import { faTrash } from '@fortawesome/free-solid-svg-icons';
...
+  } from 'mark-one';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faTrash } from '@fortawesome/free-solid-svg-icons';
+
+const BorderlessButtonRefExample = () => {
+  const ref = useRef(null);
+  const [value, setValue] = useState('');
+  const onButtonClick = () => {
+    setTimeout(() => ref.current.focus());
+  }
```
farassadek, line 105: Why not   `const onButtonClick = () => {ref.current.focus()}`?

examples/Buttons/BorderlessButton.md
```diff
@@ -79,6 +79,51 @@ import { faTrash } from '@fortawesome/free-solid-svg-icons';
...
+      Focus the Borderless Button
+    </Button>
+    <BorderlessButton
+      onClick={function() {
+        alert('You clicked the borderless button in which variant equals VARIANT.DANGER')
+      }}
+      variant={VARIANT.DANGER}
+      tabIndex={0}
+      forwardRef={ref}
+      alt="testing"
```
farassadek, line 121: I prefer `alt="testing delete"` or `alt="delete"`

examples/Buttons/Button.md
```diff
@@ -71,3 +71,45 @@ import { VARIANT } from 'mark-one';
...
+  useState,
+  useRef,
+} from 'react';
+import { VARIANT } from 'mark-one';
+
+const ButtonRefExample = () => {
+  const ref = useRef(null);
+  const onButtonClick = () => {
+    setTimeout(() => ref.current.focus());
+  }
```
farassadek, line 88: Again why not `const onButtonClick = () => {ref.current.focus()}`?

src/Buttons/__tests__/Button.test.tsx
```diff
@@ -63,4 +66,46 @@ describe('Button', function () {
...
+            >
+              Focus the Other Button
+            </Button>
+            <Button
+              onClick={(): void => {}}
+              variant={VARIANT.DANGER}
+              tabIndex={0}
+              forwardRef={ref}
+            >
+              Danger Button
```
farassadek, line 94: Trivial change : instead of `Danger Button` I prefer something like `borderless button` because the child is not always a danger (trash) button.

src/Buttons/__tests__/Button.test.tsx
```diff
@@ -63,4 +66,46 @@ describe('Button', function () {
...
+        );
+      };
+      ({ getByText } = render(
+        <ButtonRefExample />
+      ));
+    });
+    it('can be used to shift the focus to the button', async function () {
+      const initialButton = getByText('Focus the Other Button');
+      initialButton.click();
+      const dangerButton = getByText('Danger Button');
```
farassadek, line 106: Same comment about the `Danger Button`.

src/Buttons/__tests__/BorderlessButton.test.tsx
```diff
@@ -65,4 +71,47 @@ describe('Borderless Button', function () {
...
+              variant={VARIANT.PRIMARY}
+            >
+              Focus the Borderless Button
+            </Button>
+            <BorderlessButton
+              onClick={() => {}}
+              variant={VARIANT.DANGER}
+              tabIndex={0}
+              forwardRef={ref}
+              alt="Danger Button"
```
farassadek, line 98: Trivial change : instead of `Danger Button` I prefer something like `Borderless Button` because the child is not always a danger (trash) button.

src/Buttons/__tests__/BorderlessButton.test.tsx
```diff
@@ -65,4 +71,47 @@ describe('Borderless Button', function () {
...
+        );
+      };
+      ({ getByText, getByRole } = render(
+        <ButtonRefExample />
+      ));
+    });
+    it('can be used to shift the focus to the button', async function () {
+      const initialButton = getByText('Focus the Borderless Button');
+      initialButton.click();
+      const borderlessButton = getByRole('button', { name: 'Danger Button' });
```
farassadek, line 112: Same comment about the `Danger Button`.

examples/Buttons/BorderlessButton.md
natalynyu: Ok, that makes sense - updated this to "testing delete" through e8fe7be.

src/Buttons/__tests__/BorderlessButton.test.tsx
natalynyu: That's a good point - updated the naming through a5e21f1.

src/Buttons/__tests__/BorderlessButton.test.tsx
natalynyu: This was also updated through a5e21f1.

src/Buttons/__tests__/Button.test.tsx
natalynyu: This was also updated through a5e21f1.

src/Buttons/__tests__/Button.test.tsx
natalynyu: This was also updated through a5e21f1.

examples/Buttons/BorderlessButton.md
natalynyu: Ah yes, you're right that the `setTimeout` is not necessary - thanks for catching that! I removed this through 8721dbb.

examples/Buttons/Button.md
natalynyu: Thanks for catching this! I removed it through 8721dbb.

src/Theme/MarkOneWrapper.tsx
natalynyu: Some elements cannot be "tabbed" to using the tab key, but I just realized that buttons are "tab-able," so I removed the tab index from the button elements (1988d6d). However, I believe there will be some elements in the future are not included in this list of [elements that can receive focus](https://stackoverflow.com/questions/1599660/which-html-elements-can-receive-focus), so I think it makes sense to keep `tabIndex` as a property in the interface.
This is the doc I referenced for the [tabindex attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex).

farassadek: Ok make sense, thanks for the explanation and links.

src/Buttons/__tests__/BorderlessButton.test.tsx
```diff
@@ -65,4 +71,46 @@ describe('Borderless Button', function () {
...
+  context('when forwardRef prop is present', function () {
+    beforeEach(function () {
+      const ButtonRefExample = () => {
+        const ref = useRef<HTMLInputElement>(null);
+        const onButtonClick = () => {
+          setTimeout(() => {
+            if (ref.current) {
+              ref.current.focus();
+            }
+          });
```
farassadek, line 83: Do you still need the `setTimeout` in the test?

src/Buttons/__tests__/Button.test.tsx
```diff
@@ -63,4 +66,45 @@ describe('Button', function () {
       assert.strictEqual(clickSpy.callCount, 0);
     });
   });
+  context('when forwardRef prop is present', function () {
+    beforeEach(function () {
+      const ButtonRefExample = () => {
+        const ref = useRef<HTMLInputElement>(null);
+        const onButtonClick = () => {
+          setTimeout(() => {
```
farassadek, line 74: is `setTimeout` still needed in the test ?

natalynyu: Ah thanks for catching that! You're right - removed through 644e818

src/Buttons/__tests__/BorderlessButton.test.tsx
natalynyu: Also removed through 644e818, thanks!

farassadek (approved):
Looks good.

PR #81, opened 2020-11-13
Feature/224 add accessibility to the label

farassadek:
This PR related to the course planner ticket https://github.com/seas-computing/course-planner/issues/224
Based on the conversion with Vittorio, the invisible label should be accessible by the screen reader. Therefore the `dispaly: none` in the Label component in mark-one should be replaced with this css found in this url https://accessibility.18f.gov/hidden-content/

**Types of changes**

- [x] Breaking change (fix or feature that would cause existing functionality to change)

**Checklist:**
- [x] I have run eslint on the code
**Priority:**
- [x] Normal
- [ ] High

### DIFF
--- a/src/Forms/InputLabel.tsx
+++ b/src/Forms/InputLabel.tsx
@@ -65,7 +65,7 @@ const accessibility = `
   margin: -1px;
   padding: 0;
   border: 0;
-`
+`;
 
 const generateGrid = (
   labelPosition: InputLabelPosition,
@@ -104,11 +104,7 @@ const StyledInputLabel = styled.label<StyledInputLabelProps>`
 `;
 
 const StyledInputLabelText = styled.span<StyledInputLabelTextProps>`
-  ${({ isLabelVisible }) => (
-    isLabelVisible
-    ? 'display: inline;'
-    : accessibility 
-  )};
+  ${({ isLabelVisible }) => (isLabelVisible ? 'display: inline;' : accessibility)};
   grid-area: l;
   justify-self: ${({ labelPosition }) => (
     labelPosition === POSITION.TOP
--- a/src/Forms/InputLabel.tsx
+++ b/src/Forms/InputLabel.tsx
@@ -56,6 +56,17 @@ export interface InputLabelProps {
   hideError?: boolean;
 }
 
+const accessibility = `
+  position: absolute;
+  overflow: hidden;
+  clip: rect(0 0 0 0);
+  height: 1px;
+  width: 1px;
+  margin: -1px;
+  padding: 0;
+  border: 0;
+`
+
 const generateGrid = (
   labelPosition: InputLabelPosition,
   isLabelVisible: boolean
@@ -93,7 +104,11 @@ const StyledInputLabel = styled.label<StyledInputLabelProps>`
 `;
 
 const StyledInputLabelText = styled.span<StyledInputLabelTextProps>`
-  display: ${({ isLabelVisible }) => (isLabelVisible ? 'inline' : 'none')};
+  ${({ isLabelVisible }) => (
+    isLabelVisible
+    ? 'display: inline;'
+    : accessibility 
+  )};
   grid-area: l;
   justify-self: ${({ labelPosition }) => (
     labelPosition === POSITION.TOP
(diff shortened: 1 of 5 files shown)


src/Forms/InputLabel.tsx
```diff
@@ -56,6 +56,17 @@ export interface InputLabelProps {
   hideError?: boolean;
 }
 
+const accessibility = `
```
jonseitz, line 59: I think this needs a more descriptive name, maybe something like `hideFromDisplay`? Because while this is an accessibility feature it's really only dealing with one part of accessibility, so adding it doesn't necessarily make a component accessible, and the removing it doesn't make it inaccessible.

src/Forms/InputLabel.tsx
```diff
@@ -93,7 +104,7 @@ const StyledInputLabel = styled.label<StyledInputLabelProps>`
 `;
 
 const StyledInputLabelText = styled.span<StyledInputLabelTextProps>`
-  display: ${({ isLabelVisible }) => (isLabelVisible ? 'inline' : 'none')};
+  ${({ isLabelVisible }) => (isLabelVisible ? 'display: inline;' : accessibility)};
```
jonseitz, line 107: I actually don't think we need to have `display: inline` as part of this -- that's the default display value for a span anyway, and that value shouldn't affect the changes made by the other css snippet. We should be able to leave it out entirely and just have something like

```ts
  ${({ isLabelVisible }) => (isLabelVisible ? accessibility : '')};
```

src/Forms/__tests__/Dropdown.test.tsx
```diff
@@ -415,7 +415,7 @@ describe('Dropdown', function () {
     });
     it('the label value is hidden in the UI', function () {
       const style = window.getComputedStyle(getByText('semesters'));
-      strictEqual(style.display, 'none');
+      strictEqual(style.overflow, 'hidden');
```
jonseitz, line 418: Here and in the `TextInput` test, I think we want to be testing that we can get a reference to the input with `getByLabelText('InvisibleLabel')`, since that will provide better confirmation that the accessibility features are working -- I'm less concerned about testing individual css rules (and, as noted elsewhere, `window.getComputedStyle` doesn't always work correctly).

src/Forms/InputLabel.tsx
jonseitz: One other thought: Maybe we should move this into its own file (e.g. `src/utils/mixins.ts` or `src/utils/accessibility.ts`) so that we can import this into other components where need to hide text from the UI. Along the same lines as Natalyn's work in #80, this would make it easier for us to reuse common solutions across the library.

jonseitz (changes requested):
This is a good approach to fixing this, and I had just a few comments and little tweaks noted below.

src/Forms/InputLabel.tsx
farassadek: I like the idea of having this in a new component but can we open a new ticket for this because my course planning ticket is depend on this functionality.
We can split this to a new component later if that make sense to you.

src/Forms/InputLabel.tsx
farassadek: ok

src/Forms/__tests__/Dropdown.test.tsx
```diff
@@ -414,8 +420,7 @@ describe('Dropdown', function () {
       deepStrictEqual(optionsFound, optionsWithDefaults);
     });
     it('the label value is hidden in the UI', function () {
```
jonseitz, line 422: One last minor note: Can we update the name of this test to be something like "The label is still accessible in the DOM", since we're no longer testing whether it's hidden? (also applies in `TextInput.test.tsx`)

jonseitz (changes requested):
One last minor note, but otherwise this is looking good.

src/Forms/InputLabel.tsx
jonseitz: Yep, that's a good idea to wait on splitting this out.

src/Forms/__tests__/Dropdown.test.tsx
farassadek: good point, thanks.

jonseitz (approved):

PR #82, opened 2020-11-16
Add disabled property to Dropdown and onClick property to Dropdown & TextInput

natalynyu:
In the Course Planner Course Admin modal, we would like to have the "Select an existing area" dropdown to be disabled when the user is interacting with the "Create New Area" text input field, so I've exposed the `disabled` property on the `Dropdown` component. Also, it was noted in our retrospective that it would be nice to have the radio button selected when one of the elements were clicked (e.g. have the Create New Area radio button selected when the text input field is clicked), so I also included an optional `onClick` property for the `TextInput` and `Dropdown.`
image

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Addresses [#272](https://github.com/seas-computing/course-planner/issues/272)

### DIFF
--- a/src/Forms/__tests__/Dropdown.test.tsx
+++ b/src/Forms/__tests__/Dropdown.test.tsx
@@ -21,6 +21,7 @@ describe('Dropdown', function () {
   let getAllByRole: BoundFunction<AllByRole>;
   let queryByText: BoundFunction<QueryByText>;
   let changeSpy: SinonSpy;
+  let clickSpy: SinonSpy;
   const options = [
     {
       value: 'all',
@@ -418,7 +419,7 @@ describe('Dropdown', function () {
       strictEqual(style.display, 'none');
     });
   });
-  context('when disabled prop is true', function () {
+  context('when disabled prop within an option is true', function () {
     beforeEach(function () {
       changeSpy = spy();
       ({ getByText, getAllByRole } = render(
@@ -438,7 +439,7 @@ describe('Dropdown', function () {
       strictEqual(summerOption.disabled, true);
     });
   });
-  context('when disabled prop is false', function () {
+  context('when disabled prop within an option is false', function () {
     beforeEach(function () {
       changeSpy = spy();
       ({ getByText, getAllByRole } = render(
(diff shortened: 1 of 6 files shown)


src/Forms/__tests__/Dropdown.test.tsx
```diff
@@ -418,7 +419,7 @@ describe('Dropdown', function () {
       strictEqual(style.display, 'none');
     });
   });
-  context('when disabled prop is true', function () {
+  context('when disabled prop within an option is true', function () {
```
rmainseas, line 422: This test (and the one below it) aren't named correctly. The prop is on the component, not the options.

src/Forms/__tests__/TextInput.test.tsx
```diff
@@ -24,6 +24,7 @@ describe('Text input', function () {
   let getByRole: BoundFunction<GetByRole>;
```
rmainseas, line 24: Doesn't look like there's any tests for the `disabled` prop in here...

rmainseas (changes requested):
Looks great, just a couple of small things

src/Forms/__tests__/TextInput.test.tsx
natalynyu: I had them in the test file - are they not showing up? 
image

src/Forms/__tests__/Dropdown.test.tsx
natalynyu: Before this PR, there was already a way to disable the option within the dropdown, so I had just named the test "when disabled prop is true." Now we can make the entire dropdown disabled in addition to having individual options disabled, so I'm not sure how to describe the tests? Should I say for line 422 "when disabled prop for an option is true"? For the new `disabled` prop for the overall dropdown, I described it "when disabled prop is true" (line 482 of the test). 
image

src/Forms/__tests__/TextInput.test.tsx
rmainseas: They weren't, no.
image


I think it was because you didn't add this in _this_ PR GitHub just wasn't showing them. But I do see them now.

src/Forms/__tests__/Dropdown.test.tsx
rmainseas: Oh right! Sorry. I didn't realize that you could also disable `<option>`s. I thought that was just an old holdover from the previous implementation :upside_down_face: .

rmainseas (approved):
LGTM :rocket:

PR #89, opened 2021-01-26
Migrate CI/CD to GitHub Actions

jonseitz:
## Describe your changes

On the same lines as seas-computing/course-planner#296, this will reproduce our existing Travis CI scripts in GitHub Actions, adding workflows for:

1. Static code analysis
2. PR branch validity (only `develop` can merge into `main`)
3. Automatic versioning based on our [Release Management guide](https://github.com/seas-computing/dev-team-notes/blob/main/release_management.md)
4. Deployment of our styleguide to gh-pages
5. Deployment of our package to npm

Workflows 1-3 are identical to the files of the same name in course-planner.

Workflow 4 only features minor changes to the commands, to match what's defined in Mark One's `package.json`

Workflow 5 is new, but is adapted from the Docker deployment script in course-planner.

As noted in the comments on seas-computing/course-planner#296, workflows 1 and 2 will not, on their own, prevent failing PRs from being merged, so we'll need to institute some additional branch protections in the repo settings once these have been merged into `main`. In addition, we'll need to add a ~`NODE_AUTH_TOKEN`~ `NPM_TOKEN` repo secret before we can publish to npm.

# # Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes seas-computing/course-planner#295


.github/workflows/02-pr-check.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+jobs:
+  check:
+    name: Only allow develop into main
```
rmainseas, line 11: Are we likely to ever want to do hotfixes into main?

rmainseas (changes requested):
This looks great. Just one small thing.

.github/workflows/02-pr-check.yml
jonseitz: Hmmm, I suppose that is a possibility. Though, technically speaking, the status check itself doesn't stop you from merging into main, it just throws up a red flag to catch cases where we accidentally open a PR against `main` instead of `develop`, and we'd need to set additional branch protection rules on a per-repo basis to actually make this a required status check, and we could override it if we did need to merge something to main.

But yeah, I am feeling a little mixed on how useful this status check actually is, tbh. On one hand, accidentally merging into `main` is more likely to happen if we're merging `develop` more often, since they will be fewer commits different from any feature branch, and it can lead to some headaches if we end up with conflicts. But it also feels weird to have a status check that's not totally required...

What would you think about keeping it in, evaluating how useful it actually is for a month or two, then reassessing?

rmainseas (approved):

PR #93, opened 2021-02-09
Form Component and Updating Checkbox formatting

natalynyu:
## Describe your changes

This PR adds a `Form` component, which adds a margin to each of its direct children (e.g. In Course Planner, if we replace the Course Admin modal's form component using `<form>` with the `Form` component, each of the form field components will have a margin around them).

While discussing the form fields with Jonathan Seitz, we found that the checkbox in the course planner course admin modal is indented more compared to other form fields. This was because of the `padding-inline-start` property in `Fieldset,` so the value was updated depending on whether the border and legends are visible. 

Lastly, the `hideError` prop was integrated into the `Checkbox` component for formatting reasons in the same way it had been integrated with the `TextInput` and `Dropdown` components, so that the user has the option to hide the space allocated for the error message when there will not be an error.

## Types of changes

      to change)

## Checklist:

- [X] I have added JSDoc for all of my code (where applicable)
- [X] I have added tests to cover my changes.

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #274

### DIFF
--- a/src/Forms/Fieldset.tsx
+++ b/src/Forms/Fieldset.tsx
@@ -49,7 +49,11 @@ const StyledFieldset = styled.fieldset<StyledFieldsetProps>`
       : 'none'
   )};
   padding-block-start: 0.35em;
-  padding-inline-start: 0.75em;
+  padding-inline-start: ${({ isBorderVisible, isLegendVisible }): string => (
+    !isBorderVisible && !isLegendVisible
+    ? null
+    : '0.75em'
+  )};
   padding-inline-end: 0.75em;
   padding-block-end: 0.625em;
   min-inline-size: min-content;
--- a/src/Forms/Fieldset.tsx
+++ b/src/Forms/Fieldset.tsx
@@ -42,6 +42,8 @@ export interface FieldsetProps {
 interface StyledLegendProps {
   /** Specifies whether the legend of the fieldset is visible */
   isLegendVisible?: boolean;
+  /** Specifies whether the border of fieldset is visible */
+  isBorderVisible?: boolean;
 }
 
 const StyledFieldset = styled.fieldset<StyledFieldsetProps>`
@@ -65,7 +67,11 @@ const StyledFieldset = styled.fieldset<StyledFieldsetProps>`
 `;
 
 const StyledLegend = styled.legend<StyledLegendProps>`
-  padding: 0.25em;
+  padding: ${({ isBorderVisible }): string => (
+    isBorderVisible
+      ? '0.25em'
+      : null
+  )};
   position: ${({ isLegendVisible }): string => (
     isLegendVisible
       ? null
--- a/src/Forms/Fieldset.tsx
+++ b/src/Forms/Fieldset.tsx
@@ -53,8 +53,8 @@ const StyledFieldset = styled.fieldset<StyledFieldsetProps>`
   padding-block-start: 0.35em;
   padding-inline-start: ${({ isBorderVisible, isLegendVisible }): string => (
     !isBorderVisible && !isLegendVisible
-    ? null
-    : '0.75em'
+      ? null
+      : '0.75em'
   )};
   padding-inline-end: 0.75em;
   padding-block-end: 0.625em;
--- a/src/Forms/Fieldset.tsx
+++ b/src/Forms/Fieldset.tsx
@@ -6,6 +6,7 @@ import React,
   useContext,
 } from 'react';
 import styled, { ThemeContext } from 'styled-components';
+import { fromTheme } from 'Theme';
 import { NoteText } from '../Typography';
 import ValidationErrorMessage from './ValidationErrorMessage';
 import { RequiredSymbol } from './InputLabel';
@@ -59,6 +60,9 @@ const StyledFieldset = styled.fieldset<StyledFieldsetProps>`
   padding-inline-end: 0.75em;
   padding-block-end: 0.625em;
   min-inline-size: min-content;
+  & > * {
+    margin: ${fromTheme('ws', 'small')};
+  }
 `;
 
 const StyledLegend = styled.legend<StyledLegendProps>`
--- a/src/Forms/Fieldset.tsx
+++ b/src/Forms/Fieldset.tsx
@@ -15,6 +15,8 @@ interface StyledFieldsetProps {
   form?: string;
   /** Specifies whether the border of fieldset is visible */
   isBorderVisible?: boolean;
+  /** Specifies whether the legend of the fieldset is visible */
+  isLegendVisible?: boolean;
 }
 
 export interface FieldsetProps {
--- a/src/Forms/Fieldset.tsx
+++ b/src/Forms/Fieldset.tsx
@@ -51,17 +51,16 @@ const StyledFieldset = styled.fieldset<StyledFieldsetProps>`
       ? `${theme.border.light}`
       : 'none'
   )};
-  padding-block-start: 0.35em;
   padding-inline-start: ${({ isBorderVisible, isLegendVisible }): string => (
     !isBorderVisible && !isLegendVisible
       ? null
       : '0.75em'
   )};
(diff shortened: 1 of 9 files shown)


src/Forms/Form.tsx
```diff
@@ -0,0 +1,44 @@
...
+import styled, { ThemeContext } from 'styled-components';
+import { fromTheme } from '../Theme';
+
+export interface FormProps {
+  /** Text or form components to be displayed */
+  children: ReactNode;
+  /** The id of the form */
+  id?: string;
+  /** The test id of the form */
+  dataTestId?: string;
```
jonseitz, line 16: Per note on test, we should avoid using `test-id` if there's a valid `role` we can use instead

src/Forms/Form.tsx
```diff
@@ -0,0 +1,44 @@
+import React, {
+  FunctionComponent,
+  ReactElement,
+  ReactNode,
+  useContext,
+} from 'react';
+import styled, { ThemeContext } from 'styled-components';
+import { fromTheme } from '../Theme';
+
+export interface FormProps {
```
jonseitz, line 10: I think we may need to include a `name` prop for accessibility reasons

jonseitz (changes requested):
Thanks for catching all the small style tweaks in here. I do have a couple of small notes and tweaks below. I also think we might want to apply a similar margin to children of the `FieldSet`, for those cases where we might have a set of text inputs/dropdowns within a fieldset that wouldn't pick up the styling from the `Form`.

src/Forms/__tests__/Form.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+          onClick={() => {}}
+          variant={VARIANT.BASE}
+        >
+          Submit
+        </Button>
+      </Form>
+    ));
+  });
+  it('renders', function () {
+    getByTestId(formTestId);
```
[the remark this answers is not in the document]
natalynyu: Ah right - I had used the `test-id` as a final resort because the [ARIA form role docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Form_Role) were saying to add the `role` of `form` when not using `<form>` (e.g. if you're using `<div>` instead of `form`) and that using `<form>` was sufficient for letting screen readers know that it is a form. Adding a role of form does work, but I noticed when looking at the docs that we should be adding an `aria-label` to forms to distinguish their purposes - I replaced the test id property with the `label` property in 93c724b. This way, I can use `getByLabelText` in the test instead of `getByTestId.`

src/Forms/Form.tsx
natalynyu: I ended up querying for the label instead as a result of the note above.

src/Forms/Form.tsx
```diff
@@ -0,0 +1,44 @@
...
+  id?: string;
+  /** The test id of the form */
+  dataTestId?: string;
+}
+
+const StyledForm = styled.form<FormProps>`
+  & > * {
+    margin: ${fromTheme('ws', 'small')};
+  }
+  cursor: pointer;
```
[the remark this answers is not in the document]
natalynyu: @jonseitz Ah thanks for pointing this out! I'm not sure how this CSS cursor rule snuck in there - that's very odd. I removed it through 1ea2082.

src/Forms/Form.tsx
natalynyu: The [documentation on `<form>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) is saying that the `id` should be sufficient - with the `id` and the `aria-label` that was newly added, is this enough?

natalynyu: image

src/Forms/Form.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <StyledForm
+      id={id}
+      data-testid={dataTestId}
+      theme={theme}
+    >
+      {children}
+    </StyledForm>
+  );
+};
+
```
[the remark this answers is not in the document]
natalynyu: I think that now that I'm trying to specify the `aria-label` I have to define `Form` in the original way - is this correct?
image

natalynyu:
I also added a rule to add a margin to the direct children of `Fieldset` (76bae7a). Is this what you had in mind?

src/Forms/__tests__/Form.test.tsx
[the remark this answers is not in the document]
jonseitz: That link is specific to using `role="form"` on a non-`<form>` element, which we are correctly not doing. But the `ByRole` queries in `testing-library` also work with implicit accessibility roles, and because the `<form>` element does have an implicit role of `form'`, we can use `getByRole('form')` and it will correctly query for the `<form>` element and confirm that our `Form` has been rendered.

src/Forms/Form.tsx
jonseitz: Ah, you're right. That's my bad.

(I think I was confusing the `name` attribute with the "[accessible name](https://www.w3.org/TR/accname-1.1/#dfn-accessible-name)" concept)

src/Forms/Form.tsx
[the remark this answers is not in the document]
jonseitz: You could use [the `.attrs()` construction](https://styled-components.com/docs/basics#attaching-additional-props) in styled-components to do this, which would look something like:

```ts
export default styled.form.attrs(
  (props: FormProps) => ({
    'aria-label': props.label,
  })
)`
  & > * {
  }
`;
```

The value of `label` will get passed through to `aria-label`:



And that will be shown as the `name` in the Accessibility tree:

src/Forms/Fieldset.tsx
```diff
@@ -59,6 +60,9 @@ const StyledFieldset = styled.fieldset<StyledFieldsetProps>`
   padding-inline-end: 0.75em;
   padding-block-end: 0.625em;
   min-inline-size: min-content;
+  & > * {
+    margin: ${fromTheme('ws', 'small')};
```
jonseitz, line 64: I don't necessarily know if there's a right answerto this, but with the margin added to the child component do you think we still need the padding inside the `FieldSet` as well? 

I've looked at it both ways a couple times and I _think_ I prefer it without the padding on the top and bottom of the fieldset, but I think it might make since to keep the padding on the left and right. What do you think?

jonseitz (changes requested):
Responded on your comments below, and added a note about the fieldset changes.

src/Forms/__tests__/Form.test.tsx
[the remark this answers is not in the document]
natalynyu: Ok thanks! This was my original approach (to do `getByRole`), but I must have done it wrong initially, because I was getting an error saying that the role 'form' couldn't be found.. so I probably had a typo or something the first time around, which sent me down this rabbit hole! I updated the test to use `getByRole` in commit aadb274.

src/Forms/Fieldset.tsx
natalynyu: I think that we can remove the padding on the top and bottom too.

However, I noticed one thing with the margin which was that this new margin rule was making it so that the checkbox, which needs to be in a fieldset for accessibility reasons, is indented.
Screen Shot 2021-02-11 at 4 10 27 PM
 I updated the margin rule to only add padding to the `margin-top` and `margin-bottom`. What do you think?

jonseitz: Ah, now that you point that out it's super noticable. Good eye :eye: 

I think these latest changes look great, and that combination of margins and paddings should be all set.

src/Forms/Form.tsx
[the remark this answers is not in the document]
natalynyu: Thanks for showing me this! I updated the Form component to use `.attrs` instead in 1a467d9.

jonseitz (approved):
Looks great! This is all set!

PR #104, opened 2021-06-07
Button Dropdown Menu Component 

natalynyu:
## Describe your changes

This PR adds the `ButtonDropdown` component. When the button, which must be a Font Awesome icon, is clicked, a menu of options expands. If the button is initially clicked and clicked again, the menu is minimized. If the user clicks outside the menu, the menu also minimizes. Lastly, when the user selects an item in the menu, the `onClick` instructions provided are executed and then the menu is minimized. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes [#343](https://github.com/seas-computing/course-planner/issues/343)

### DIFF
--- a/src/Buttons/ButtonDropdown.tsx
+++ b/src/Buttons/ButtonDropdown.tsx
@@ -20,14 +20,6 @@ interface DropdownOptionProps {
   value: string;
 }
 
-export interface MenuProps {
-  /**
-   * Whether or not the menu should be visible.
-   * This will be controlled by the parent component, likely via useState.
-   */
-  isMenuVisible: boolean;
-}
-
 export interface ButtonDropdownProps {
   /** The id of the button */
   id?: string;
@@ -35,9 +27,7 @@ export interface ButtonDropdownProps {
   children: ReactNode;
   /** Function to call on click of a selection within the dropdown */
   onChange: (arg0: string) => void;
-  /**
-   * An array of string that outline the choices in the dropdown
-   */
+  /** An array of string that outline the choices in the dropdown */
   options: Array<DropdownOptionProps>;
   /** Specifies the alt text for screen readers */
   alt: string;
@@ -60,7 +50,7 @@ const StyledButtonDropdown = styled.button`
   }
 `;
 
-const StyledMenu = styled.div<MenuProps>`
+const StyledMenu = styled.div`
   background: ${fromTheme('color', 'background', 'light')};
   border: ${fromTheme('border', 'hairline')};
   min-height: 3em;
@@ -139,7 +129,7 @@ const ButtonDropdown: FunctionComponent<ButtonDropdownProps> = (props)
         { children }
       </StyledButtonDropdown>
       {isMenuVisible && (
-        <StyledMenu isMenuVisible={isMenuVisible}>
+        <StyledMenu>
           <StyledMenuList>
             {options.map((option) => (
               <StyledMenuListItem
--- a/src/Buttons/ButtonDropdown.tsx
+++ b/src/Buttons/ButtonDropdown.tsx
@@ -1,0 +1,176 @@
+import { FontAwesomeIconProps } from '@fortawesome/react-fontawesome';
+import { Dropdown } from 'Forms';
+import React, {
+  FunctionComponent,
+  ReactElement,
+  useContext,
+  ReactNode,
+  MouseEventHandler,
+  useRef,
+  ChangeEventHandler,
+} from 'react';
+import styled, { ThemeContext } from 'styled-components';
+import {
+  fromTheme,
+  MarkOneProps,
+} from '../Theme';
+
+interface DropdownOptionProps {
+  /** The label of the dropdown option */
+  label: string;
+  /** The value of the dropdown option */
+  value: string;
+}
+
+export interface MenuProps {
+  /**
+   * Whether or not the menu should be visible.
+   * This will be controlled by the parent component, likely via useState.
+   */
+  isMenuVisible: boolean;
+  /**
+   * The content to be displayed within the modal
+   */
+  children: ReactNode;
+}
+
+export interface ButtonDropdownProps extends MarkOneProps<HTMLButtonElement> {
+  /** The id of the button */
+  id?: string;
+  /** Specifies the Font Awesome Icon(s) */
+  children: ReactElement<FontAwesomeIconProps>;
+  /** Function to call on click of the dropdown button */
+  onDropdownClick: MouseEventHandler;
+  /** Function to call on click of a selection within the dropdown */
+  onChange: ChangeEventHandler;
+  /**
+   * Whether or not the menu should be visible.
+   * This will be controlled by the parent component, likely via useState.
+   */
+  isMenuVisible: boolean;
+  /**
+   * An array of string that outline the choices in the dropdown
+   */
+  options: Array<DropdownOptionProps>;
+}
+
+const StyledButtonDropdown = styled.button<ButtonDropdownProps>`
+  background: ${fromTheme('color', 'background', 'medium')};
+  border: none;
+  cursor: pointer;
+  display: inline-block;
+  font-size: ${fromTheme('font', 'body', 'size')};
+  font-weight: ${fromTheme('font', 'body', 'weight')};
+  padding: ${({ theme }) => (`${theme.ws.xsmall} ${theme.ws.small}`)};
+  &:hover {
+    background: ${fromTheme('color', 'background', 'dark')};
+    border: none;
+    color: ${fromTheme('color', 'text', 'base')};
+  }
+`;
+
+const StyledMenu = styled.div<MenuProps>`
+  background: ${fromTheme('color', 'background', 'light')};
+  border: ${fromTheme('border', 'hairline')};
+  min-height: 3em;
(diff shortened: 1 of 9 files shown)


src/Buttons/ButtonDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+        ref={forwardRef}
+        aria-label={alt}
+      >
+        { children }
+      </StyledButtonDropdown>
+      {isMenuVisible && (
+        <>
+          <StyledMenu isMenuVisible={isMenuVisible}>
+            <StyledMenuList>
+              {options.map((option) => (
```
jonseitz, line 145: Just thinking about this some more, I feel like this is a bit close to how we do our `TabList` than we do the `Dropdown`. And with the tabs, we expose both the `TabList` and the `TabListItem` and let the user compose them however they want. I wonder if something more like that would make more sense here, where we could expose `ButtonDropdownMenu` and `ButtonDropdownMenuItem` components and let the application handle looping through them. My thought there is that:

1) We don't have to be as prescriptive about the `onClick` handler or the shape of the `options`, so in the case of our time picker we could, in course-planner, have something like `{ label: '1:00 PM - 2:00PM', startTime: '13:00:00', endTime: '14:00:00 }` and handle those separately.

2) It would potentially let us put non-clickable items (i.e. headers) in the menu.

I may be overthinking things at this point, but wanted to put it out there in case it make sense to you too.

jonseitz, line 145: Thinking about this some more, I'm starting to feel like it would make more sense to change `ButtonDropdown` to `ButtonDropdownMenu` and separate export a `ButtonDropdownMenuItem` component. Then we could let the consuming application compose the menu however it wants by making the contents of the `StyledMenuList` be the children of `ButtonDropdownMenu`, similar to how the `TabList` and `TabListItem` work. (In that case, we'd want to add a `label` prop to that `ButtonDropdownMenu` to set the content of the `Button`). My thinking here is:

1.) We wouldn't need to be as limited in the shape of `options` and how the `onChange` handler gets called, since that would be done as part of the implementation of `ButtonDropdownMenuItem`. For instance, in the course-planner, we could have `startTime` and `endTime` props that could be handled separately. So something like:

```ts
// in course planner

<ButtonDropdownMenu label={<FontAwesomeIcon />}>
  {[
    { 
      label: '1:00 PM - 2:00 PM',
      startTime: '13:00:00',
      endTime: '14:00:00',
    }, 
    ...rest,
  ].map(({label, startTime, endTime}) => (
    <ButtonDropdownMenuItem
      onClick={() => {
        // state setters from parent
        setStartTime(startTime);
        setEndTime(endTime);
      })
    >
      {label}
    <ButtonDropdownMenuItem>
  ))}
</ButtonDropdownMenu>

2.) We could potentially add non-button items to the list, e.g. headings or links

I may be trying to over-optimize, but that feels like a more react-y way of doing things. What do you think?

File: src/Buttons/ButtonDropdown.tsx [hunk 2]
```diff
@@ -0,0 +1,166 @@
...
+                <StyledMenuListItem
+                  key={option.value}
+                  onClick={() => {
+                    onChange(option.value);
+                    setMenuVisible(false);
```
jonseitz, line 150: This call to `setMenuVisible` isn't actually needed, because the `body`-level listener is also going to capture this click event and set the visibility to `false`.

src/Buttons/ButtonDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+  transform: translate3d(0em, 0.1em, 0em);
+  z-index: 100;
+`;
+
+const StyledMenuList = styled.ul`
+  list-style-type: none;
+`;
+
+const StyledMenuListItem = styled.li`
+  padding: .25em 1em;
```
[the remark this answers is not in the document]
natalynyu: Definitely, updated this in 4bb59a9.

src/Buttons/ButtonDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+   * This will be controlled by the parent component, likely via useState.
+   */
+  isMenuVisible: boolean;
+}
+
+export interface ButtonDropdownProps {
+  /** The id of the button */
+  id?: string;
+  /** Specifies the Font Awesome Icon */
+  children: ReactElement<FontAwesomeIconProps>;
```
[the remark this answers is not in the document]
natalynyu: Good idea - this was updated in 1684c81.

src/Buttons/ButtonDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+          setMenuVisible(!isMenuVisible);
+        }}
+        theme={theme}
+        ref={forwardRef}
+        aria-label={alt}
+      >
+        { children }
+      </StyledButtonDropdown>
+      {isMenuVisible && (
+        <>
```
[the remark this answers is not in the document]
natalynyu: Oops! This was fixed in dc298f8.

src/Buttons/ButtonDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+          <StyledMenu isMenuVisible={isMenuVisible}>
```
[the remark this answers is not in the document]
natalynyu: Thanks for catching this! This was part of an earlier implementation that I should've removed. Removed in 6d70539.

src/Buttons/ButtonDropdown.tsx
natalynyu: Good point. This was removed in 84216e6.

src/Buttons/ButtonDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+   * An array of string that outline the choices in the dropdown
+   */
+  options: Array<DropdownOptionProps>;
+  /** Specifies the alt text for screen readers */
+  alt: string;
+  /** Specifies the ref of the element */
+  forwardRef?: Ref<HTMLButtonElement>;
+}
+
+const StyledButtonDropdown = styled.button`
```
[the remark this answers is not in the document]
natalynyu: That definitely makes more sense. The component was updated to use our `Button` component in fb14733.

src/Buttons/ButtonDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+        <>
+          <StyledMenu isMenuVisible={isMenuVisible}>
+            <StyledMenuList>
+              {options.map((option) => (
+                <StyledMenuListItem
+                  key={option.value}
+                  onClick={() => {
+                    onChange(option.value);
+                    setMenuVisible(false);
+                  }}
```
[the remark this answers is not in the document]
natalynyu: Ok that makes sense. I moved the listener to the button instead through c7e9f79 and also updated the CSS so that the button would fill the space of the list item.

src/Buttons/ButtonDropdown.tsx
natalynyu: Thanks for your suggestion - I like the idea of giving users more flexibility. I made the update through commit 15db8d9. I think I executed this in the way you had this in mind, but let me know if you had something else in mind!

natalynyu: Thanks for outlining how the code would look in course planner - that's a huge help for me! I think the changes were addressed in commit 15db8d9.

src/Buttons/ButtonDropdownMenuItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+const ButtonDropdownMenuItem
+: FunctionComponent<ButtonDropdownMenuItemProps> = (props): ReactElement => {
+  const {
+    onClick,
+    label,
+    value,
+  } = props;
+  return (
+    <StyledMenuListItem
+      value={value}
```
jonseitz, line 45: Is `value` necessary here?  It doesn't seem like it's actually used...

src/Buttons/ButtonDropdownMenuItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+    >
+      <StyledMenuButton
+        onClick={onClick}
+      >
+        {label}
```
jonseitz, line 50: Instead of exposing this as `label`, what if we just used `children` for this?

jonseitz (changes requested):
Two small notes on the `ButtonDropdownMenuItem`, but otherwise this looks great!

src/Buttons/ButtonDropdown.tsx
jonseitz: Weird, I thought this comment had disappeared. Sorry to double up on this.

src/Buttons/ButtonDropdownMenuItem.tsx
natalynyu: Ah you're right thanks!

src/Buttons/ButtonDropdownMenuItem.tsx
natalynyu: Sure, I replaced `label` with children in 2557912.

jonseitz (approved):
Looks great! This is approved!

PR #119, opened 2021-09-29
Create generic list components

jonseitz:
These components are variations of the `TableCellList` and `TableCellListItem`, with an "ordered" prop added to the List component to switch between a `<ul>` and `<ol>` element. This doesn't affect the appearance of the list, but it still important semantically.

Additionally, the borders between the list children have been moved to the parent List element, which provides just a bit more flexibility if we ever want to create multiple list components.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes seas-computing/course-planner#407

### DIFF
--- a/src/Lists/List.tsx
+++ b/src/Lists/List.tsx
@@ -1,0 +1,42 @@
+import styled from 'styled-components';
+import ListItem from './ListItem';
+import { fromTheme } from '../Theme';
+
+export interface ListProps {
+  /**
+   * Should only contain ListItems
+   */
+  children: ListItem | ListItem[];
+  /**
+  * Whether this list is ordered (true) or unordered (false)
+  */
+  ordered?: boolean;
+}
+
+/**
+* Renders either an ordered or unordered list component, depending on the value
+* of the "ordered" prop. Both list types are styled identically, but the
+* difference is important semantically
+*/
+
+const List = styled.ol.attrs(
+  ({ ordered }: ListProps) => (
+    {
+      as: ordered ? 'ol' : 'ul',
+    }
+  )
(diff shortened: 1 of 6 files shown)


src/Lists/List.tsx
```diff
@@ -0,0 +1,42 @@
...
+      as: ordered ? 'ol' : 'ul',
+    }
+  )
+)<ListProps>`
+  list-style: none;
+  > li {
+    border-top: ${fromTheme('border', 'hairline')};
+    &:first-child {
+      border-top: none;
+    }
```
natalynyu, line 34: I'm wondering how much flexibility we should give to the user in terms of styling by allowing them to specify whether they want the default list styles to show and whether they want the separating border via boolean props for each of these two cases. I suppose we could come back to expand the props that could be provided and offer more flexibility in the future too, but maybe it would be easier to do this upfront.

natalynyu (commented):
Looks great! I just had a question regarding flexibility of styling.

src/Lists/List.tsx
jonseitz: In principle, yes, we want to be flexible, but I also think there's some value to enforcing consistency through the design system (back in my magazine days I used to lose my mind over how many different ways people styled lists in articles). So I'd rather avoid adding options to cover hypothetical cases and create something that works for what we need, since we're the primary users of mark-one.

natalynyu: Ok good point - we can always come back and adjust if needed.

natalynyu (approved):
LGTM! Approved.

PR #120, opened 2021-10-06
Add 'role' and 'aria-live' properties to ModalMessage

natalynyu:
In most cases, I would anticipate the `role` and `aria-live` values would be 'alert' and 'assertive, respectively. Since this may not always be the case, I added the props `role` and `ariaLive` so that the user can change the values but also added default values for both of these properties.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #374

### DIFF
--- a/src/Modals/ModalMessage.tsx
+++ b/src/Modals/ModalMessage.tsx
@@ -14,7 +14,7 @@ export interface ModalMessageProps {
   /** The aria role of the message displayed */
   role?: string;
   /** The value of the aria-live property */
-  ariaLive?: ARIA_LIVE_VARIANT;
+  ariaLive?: string;
 }
 
 const StyledModalMessage = styled.span<ModalMessageProps>`
--- a/src/Modals/ModalMessage.tsx
+++ b/src/Modals/ModalMessage.tsx
@@ -12,9 +12,9 @@ export interface ModalMessageProps {
   /** Allows you to pass in a variant property from the VARIANT enum */
   variant: VARIANT;
   /** The aria role of the message displayed */
-  role: string;
+  role?: string;
   /** The value of the aria-live property */
-  ariaLive: string;
+  ariaLive?: ARIA_LIVE_VARIANT;
 }
 
 const StyledModalMessage = styled.span<ModalMessageProps>`
--- a/src/Modals/ModalMessage.tsx
+++ b/src/Modals/ModalMessage.tsx
@@ -1,8 +1,9 @@
+import { ARIA_LIVE_VARIANT } from 'const';
 import React, {
   FunctionComponent, ReactElement, useContext,
 } from 'react';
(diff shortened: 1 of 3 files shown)


src/Modals/ModalMessage.tsx
```diff
@@ -11,6 +11,10 @@ export interface ModalMessageProps {
   children: string;
   /** Allows you to pass in a variant property from the VARIANT enum */
   variant: VARIANT;
+  /** The aria role of the message displayed */
```
jonseitz, line 14: Can we add a brief note about what value(s) can/should be used here, and what the default value is?

src/Theme/utils.ts
```diff
@@ -123,3 +123,9 @@ export enum TEXT_VARIANT {
   NEGATIVE = 'negative',
   BASE = 'base',
 }
+
+export enum ARIA_LIVE_VARIANT {
```
jonseitz, line 127: I don't think `Theme/utils` is the right place for this, since it's not really related to the theme. What would you think about adding a `const.ts` file at the top level for this, and any future non-theme constants?

src/Theme/utils.ts
```diff
@@ -123,3 +123,9 @@ export enum TEXT_VARIANT {
...
+  OFF = 'off',
+  ASSERTIVE = 'assertive',
+  POLITE = 'polite',
+}
```
jonseitz, line 131: Can we also provide documentation on what the different values here signify? i.e "Use `ASSERTIVE` in this case. Use `POLITE` in this one."

src/Modals/ModalMessage.tsx
```diff
@@ -27,13 +31,17 @@ const ModalMessage: FunctionComponent<ModalMessageProps> = (props)
     id,
     children,
     variant,
+    role,
+    ariaLive,
   } = props;
   const theme = useContext(ThemeContext);
   return (
     <StyledModalMessage
```
jonseitz, line 39: This is also not something to fix here, but one downside of using `FunctionComponent`s that wrap `StyledComponent`s is that we lose the implicit props that would already exist on the `StyledComponent`, including things like `role` and `aria-live`. So if this were just doing `export default StyledModalMessage;`, we wouldn't need to specify every html prop that we wanted to use (or things like `ref` values).

Again, not something that needs to change here, but in the future we may end refactoring a few of these to give us more flexibility.

jonseitz (changes requested):
A few informational notes below, and then just a couple places where we need some additional documentation.

src/Modals/ModalMessage.tsx
natalynyu: Definitely. This was added in 8625dd3.

jonseitz (approved):
Looks great! Thanks for adding those documentation notes!

PR #127, opened 2022-02-01
Text area component

rmainseas:
## Describe your changes
Implements a multi-line text input component (`textarea` under the hood) to fulfil https://github.com/seas-computing/course-planner/issues/452

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes https://github.com/seas-computing/course-planner/issues/452

### DIFF
--- a/examples/Forms/MultiLineTextInput.md
+++ b/examples/Forms/MultiLineTextInput.md
@@ -144,6 +144,5 @@ const [value, setValue] = useState('');
     label="Course Notes"
     placeholder="Your text here"
     onChange={() => {}}
-    forwardRef={ref}
   />
 ```
--- a/examples/Forms/MultiLineTextInput.md
+++ b/examples/Forms/MultiLineTextInput.md
@@ -124,7 +124,9 @@ const onButtonClick = () => {
     name="course-notes"
     label="Course Notes"
     placeholder="Your text here"
-    onChange={() => {}}
+    onChange={(event) => {
+      setValue(event.target.value);
+    }}
     forwardRef={ref}
   />
 </>
@@ -138,11 +140,13 @@ import { MultiLineTextInput, Button, VARIANT } from 'mark-one';
 import { useState, useRef } from 'react';
 
 const [value, setValue] = useState('');
-  <MultiLineTextInput
-    id="text input"
-    name="course-notes"
-    label="Course Notes"
-    placeholder="Your text here"
-    onChange={() => {}}
-  />
+<MultiLineTextInput
+  id="text input"
+  name="course-notes"
+  label="Course Notes"
+  placeholder="Your text here"
+  onChange={(event) => {
+    setValue(event.target.value);
+  }}
+/>
 ```
--- a/examples/Forms/MultiLineTextInput.md
+++ b/examples/Forms/MultiLineTextInput.md
@@ -9,6 +9,7 @@ const [value, setValue] = useState('CS50 is an on-campus and online introductory
 <MultiLineTextInput
   value={value}
   id="course-notes"
+  name="course-notes"
   label="Course notes"
   onChange={(event) => {
     setValue(event.target.value);
@@ -26,6 +27,7 @@ const [value, setValue] = useState('');
 <MultiLineTextInput
   value={value}
   id="course-notes"
+  name="course-notes"
   label="Course notes"
   errorMessage="Please provide a value for this field"
   onChange={(event) => {
@@ -44,6 +46,7 @@ const [value, setValue] = useState('');
 <MultiLineTextInput
   value={value}
   id="course-notes"
+  name="course-notes"
   label="Course notes"
   isLabelVisible={false}
   onChange={(event) => {
@@ -63,6 +66,7 @@ const [value, setValue] = useState('');
 <MultiLineTextInput
   value={value}
   id="course-notes"
+  name="course-notes"
   label="Course notes"
   isRequired
   onChange={(event) => {
@@ -81,6 +85,7 @@ const [value, setValue] = useState('Some text that you can\'t change');
 <MultiLineTextInput
   value={value}
   id="course-notes"
+  name="course-notes"
   label="Course notes"
   isDisabled
   onChange={(event) => {
@@ -112,6 +117,7 @@ const onButtonClick = () => {
   </Button>
   <MultiLineTextInput
     id="text input"
+    name="course-notes"
     label="Course Notes"
     onChange={() => {}}
     forwardRef={ref}
--- a/examples/Forms/MultiLineTextInput.md
+++ b/examples/Forms/MultiLineTextInput.md
@@ -64,7 +64,6 @@ const [value, setValue] = useState('');
   value={value}
   id="course-notes"
   label="Course notes"
-  isLabelVisible={false}
   isRequired
   onChange={(event) => {
     setValue(event.target.value);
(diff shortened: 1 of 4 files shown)


examples/Forms/MultiLineTextInput.md
```diff
@@ -0,0 +1,159 @@
...
+  id="course-notes"
+  label="Course notes"
+  isLabelVisible={false}
+  onChange={(event) => {
+    setValue(event.target.value);
+  }}
+/>
+```
+
+Multi-line text box with a hidden label
```
natalynyu, line 75: Looks like there is a duplicated example of a hidden label `MultiLineTextInput`.

examples/Forms/MultiLineTextInput.md
```diff
@@ -0,0 +1,159 @@
...
+  value={value}
+  id="course-notes"
+  label="Course notes"
+  isLabelVisible={false}
+  isRequired
+  onChange={(event) => {
+    setValue(event.target.value);
+  }}
+/>
+```
```
natalynyu, line 112: For the `TextInput,` when isRequired is true, there is a red asterisk that appears next to the label. I think it may be good to update this example to show that by making `isLabelVisible` true instead of false.

src/Forms/MultiLineTextInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+  /** The value of the multi-line text area */
+  value?: string;
+  /** Function to call on change event */
+  onChange: ChangeEventHandler;
+  /** Used to control the suppression of the error message value */
+  hideError?: boolean;
+  /** Marks the field as required */
+  isRequired?: boolean;
+  /** The label to display beside this multi-line input field(if any) */
+  label?: string;
```
natalynyu, line 27: I believe the `label` should be a required prop for accessibility.

natalynyu: Additionally I think that we need the `name` property since this will be part of a form, and in the case that there is no label, I think it could be useful to also have the `placeholder` property exposed:
image

src/Forms/MultiLineTextInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+  /** Wether to show or hide the label */
```
natalynyu, line 28: Just spelling (Whether)

examples/Forms/MultiLineTextInput.md
```diff
@@ -0,0 +1,159 @@
...
+  id="course-notes"
+  label="Course notes"
+  errorMessage="Please provide a value for this field"
+  onChange={(event) => {
+    setValue(event.target.value);
+  }}
+/>
+```
+
+Multi-line text box with a hidden error
```
natalynyu, line 37: What would be the use case when we want to hide an error?

natalynyu (changes requested):
This looks great! Style wise, I'm wondering is whether the label and text area field need to be centered vertically like the `TextInput` field is with its label (maybe a good question to ask Vittorio?), and I'm also wondering if there is a need for the `POSITION.TOP` label position. If there is, it may be good to include an example of what that looks like in the markdown examples.

examples/Forms/MultiLineTextInput.md
rmainseas: Actually not sure. I added it for consistency with `TextInput`. It might make sense for it to be removed, though. Done in 9bcb05883e5eaf6d6aecb5b8a852a90c0523c85f

src/Forms/MultiLineTextInput.tsx
rmainseas: Good point. Fixed in 81a008f.

examples/Forms/MultiLineTextInput.md
rmainseas: Whops! Copy/paste strikes again!

Fixed in 76561f8

examples/Forms/MultiLineTextInput.md
rmainseas: You're right. Actually I just removed that prop altogether from this example, since it defaults to `true` anyway in 630d720

src/Forms/MultiLineTextInput.tsx
rmainseas: Yes, it should. Fixed in 0ff2505e7ed4b8513fcfadb85131f5c4d3a1afb6

rmainseas: > Additionally I think that we need the `name` property since this will be part of a form, and in the case that there is no label, I think it could be useful to also have the `placeholder` property exposed: image

D'oh. I knew I'd forgotten something. Yeah, we need the `name` prop. Added that in 9b57d72


As for the `placeholder` prop - I also agree. Fixed that in be3eed4b55c92ff4810b51159d7e2c52509e5fa8

examples/Forms/MultiLineTextInput.md
```diff
@@ -0,0 +1,149 @@
...
+
+const [value, setValue] = useState('');
+  <MultiLineTextInput
+    id="text input"
+    name="course-notes"
+    label="Course Notes"
+    placeholder="Your text here"
+    onChange={() => {}}
+    forwardRef={ref}
+  />
```
natalynyu, line 148: We could either remove or define the ref for this example:
image

src/Forms/MultiLineTextInput.tsx
```diff
@@ -0,0 +1,112 @@
...
+  labelPosition?: InputLabelPosition;
+  /** Specifies the ref of the text input */
+  forwardRef?: Ref<HTMLTextAreaElement>;
+  /** The value to set for the `name` prop of the underlying `textarea` element */
+  name: string;
+  /** Provides a placeholder value for the `textarea` */
+  placeholder: string;
+}
+
+export const StyledTextArea = styled.textarea<MultiLineTextInputProps>`
```
natalynyu, line 40: Do you know if it is expected that the label and text area need to be aligned vertically like they are for the `TextInput` component? Maybe this is something to ask Vittorio? Although, I think our currently existing use case for the text area component is the notes modal in Course Planner, and that doesn't include a label.

natalynyu: I also wonder if we should have a visible scrollbar when the text takes up more room than the size of text area box. Maybe that's another question for Vittorio?

natalynyu (changes requested):
Just a few small comments.

src/Forms/MultiLineTextInput.tsx
rmainseas: @Vittorio2015 may have some thoughts here, but I thought this was the default behavior for textarea elements

examples/Forms/MultiLineTextInput.md
rmainseas: D'oh! Probably makes sense to remove it. Fixed that in 96e6a04

src/Forms/MultiLineTextInput.tsx
rmainseas: > Do you know if it is expected that the label and text area need to be aligned vertically like they are for the `TextInput` component? Maybe this is something to ask Vittorio? Although, I think our currently existing use case for the text area component is the notes modal in Course Planner, and that doesn't include a label.

This might be something to chat about in the 9:30 standup?

> I also wonder if we should have a visible scrollbar when the text takes up more room than the size of text area box. Maybe that's another question for Vittorio?

@Vittorio2015 may have some thoughts here, but I thought this was the default behavior for textarea elements

natalynyu (approved):
Everything looks good to me. Approved!

PR #129, opened 2022-05-23
Popover component

jonseitz:
This renders a small informational pop-up box whose visibility can be
controlled via an `isVisible` prop. When set to visible, it will appear
directly above and to the right of its location in the DOM. Optional
`xOffset` and `yOffset` parameters can be used with positive and negative
values to move the popup into different positions.

I did explore adding logic to prevent the popup from appearing outside
the visible portion of the page, but ultimately decided that was adding
unnecessary complexity. That feature could be enabled at implementation
time using a callback Ref, as in the React Docs example for measuring a
DOM node[1]

[1]: https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #495

### DIFF
--- a/src/Popover/Popover.tsx
+++ b/src/Popover/Popover.tsx
@@ -1,0 +1,145 @@
+import React, {
+  Ref,
+  FunctionComponent,
+  ReactNode,
+} from 'react';
+import styled from 'styled-components';
+import { fromTheme } from 'mark-one';
+
+/**
+ * Expected props for a Popover Component
+ */
+
+interface PopoverProps {
+  /**
+   * The content that will appear inside the body of the popover
+   */
+  children: ReactNode;
+  /**
+   * An optional passthrough className to allow restyling the Popover
+   * Will be applied to the PopoverWrapper, allowing for cascading styles to be
+   * applied to the internal components
+   */
+  className?: string;
+  /**
+   * An optional forwarded Ref that will be attached to the PopoverBody, e.g.
+   * for focusing the content
+   */
+  forwardRef?: Ref<HTMLDivElement>;
+  /**
+   * Control whether the Popover should be displayed on the page.
+   */
+  isVisible: boolean;
+  /**
+   * The optional content that will appear as the heading of the Popover
+   */
+  title?: ReactNode;
+  /**
+   * The Optional  offset that should be applied to the left side of the the
+   * Popover element, relative to its placement in the DOM.
+   * Should specified in pi, em, rem, % or 0.
+   */
+  xOffset?: string;
+  /**
+   * The Optional  offset that should be applied to the bottom of the the
+   * Popover element, relative to its placement in the DOM.
+   * Should specified in pi, em, rem, % or 0.
+   */
+  yOffset?: string;
+}
+
+/**
+ * Sets the amount of offset that should be applied to the Popover element
+ */
+interface PopoverWrapperProps {
+  /**
+   * Sets the position of the left edge of the Popover
+   */
+  left: PopoverProps['xOffset'];
+  /**
+   * Sets the position of the bottom edge of the Popover
+   */
+  bottom: PopoverProps['yOffset'];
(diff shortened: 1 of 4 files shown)


src/Popover/Popover.tsx
```diff
@@ -0,0 +1,145 @@
...
+}) => (
+  <PopoverPlaceholder>
+    {isVisible
+      ? (
+        <PopoverWrapper
+          className={className}
+          left={xOffset}
+          bottom={yOffset}
+        >
+          {title ? <strong>{title}</strong> : null}
```
rmainseas, line 128: I wonder if `strong` is the most semantically appropriate element for a title here? Would a `h5` or` h6` or something be better? [Bootstrap's example](https://getbootstrap.com/docs/4.0/components/popovers/) uses an `h3` for this....(though admittedly, that's likely inserted by the component consumer, rather than the component itself)

src/Popover/Popover.tsx
```diff
@@ -0,0 +1,145 @@
...
+          <p ref={forwardRef}>{children}</p>
```
rmainseas, line 129: Is there ever a time where we might want something other than text as `children`? Especially since `children` is of type `ReactNode`, not `string`....

src/Popover/Popover.tsx
```diff
@@ -0,0 +1,145 @@
...
+   * Sets the position of the left edge of the Popover
+   */
+  left: PopoverProps['xOffset'];
+  /**
+   * Sets the position of the bottom edge of the Popover
+   */
+  bottom: PopoverProps['yOffset'];
+}
+
+const PopoverPlaceholder = styled.div`
```
rmainseas, line 65: What's the purpose of this element? Is it solely to provide an "anchor point" (for want of a better term) to apply the `className` prop to?

rmainseas (changes requested):
This looks great! Just a few quick notes.

src/Popover/Popover.tsx
jonseitz: I did consider using an `h#` component here, but my concern is that there's no way to know where in the page hierarchy this would be inserted, and I figured it's worse semantically to use headings out of order, and because these are generally going to be temporary elements, it could cause problems for a generated TOC to have these as headings that appear and disappear.

src/Popover/Popover.tsx
jonseitz: Sure, we could potentially have something like a link or a button in there. Or a FontAwesome Icon.

src/Popover/Popover.tsx
jonseitz: Yeah, it's basically an anchor that makes the {x|y}Offset props move the Popover from its place in the DOM. Basically the `position: absolute` setting on the `PopoverWrapper` requires a positioned parent element, otherwise it applies the positional properties (`left`, `bottom`) from the body of the page instead.

src/Popover/Popover.tsx
rmainseas: That makes a ton of sense.

src/Popover/Popover.tsx
rmainseas: Gotcha. Thanks for the explanation!

src/Popover/Popover.tsx
rmainseas: Right, does it perhaps make sense to not wrap this in a `<p>` to allow for better semantics?

jonseitz: OK, I can see where that makes sense. I decided to replace both the `<strong>` and `<p>` with `<div>`s in f693d103 and db73e73b, which should improve both semantics and flexibility.

rmainseas (approved):
Looks great! Merge away!

PR #131, opened 2022-06-02
Size Popover to content

jonseitz:
Just having a max-width was causing weird issues when I tried to add
longer text into the body of the popover in Course Planner. The fastest
fix I found was to also set a width equal to max-content, which makes it
use as much space as it actually needs, up to that 10em max-width.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High


natalynyu (approved):
Approved!

PR #136, opened 2022-07-29
Bug fix: Modal Closing Behavior

natalynyu:
This PR updates the `Modal` component such that the modal does not close when there is a `mouseDown` event on the modal and a `mouseUp` event on the modal backdrop. The `closeHandler` is only called when the `mouseDown` and `mouseUp` event both occur on the backdrop.

Previously, `event.stopPropagation` was being called when there was an `onClick` event on the modal. In the case where there was only a `mouseDown` on the modal, the propagation was not being stopped from bubbling up to the backdrop, so when the `mouseUp` event happened on the backdrop, it seems that it was as if a full click event happened on the backdrop.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes [#538](https://github.com/seas-computing/course-planner/issues/538)

### DIFF
--- a/src/Modals/Modal.tsx
+++ b/src/Modals/Modal.tsx
@@ -171,6 +171,8 @@ const Modal: FunctionComponent<ModalProps> = ({
     };
   }, [finalForwardRef, focusables, isVisible]);
 
+  let mouseDownOnModal = false;
+
   return createPortal((
     <CSSTransition
       appear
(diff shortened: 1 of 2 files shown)


rmainseas (approved):
This looks good to me. Approved.

PR #137, opened 2022-08-03
Remove unnecessary dependencies

jonseitz:
The mark-one npm package is bringing in some unnecessary dependencies, one of which, `canvas`, doesn't have a proper `arm64` build so it requires a somewhat onerous workaround to install on M1 Macbooks (see seas-computing/course-planner#455). Removing these should slim down the installation and make life much easier for the downstream projects that use `mark-one`, like the course planner.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality
      to change)

## Checklist:

- [ ] I have run `eslint` on the code
- [ ] I have added JSDoc for all of my code (where applicable)
- [ ] I have added tests to cover my changes.

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

seas-computing/course-planner#455

PR #138, opened 2022-08-08
Refactor fix for Modal background click behavior

jonseitz:
## Describe your changes

While trying to test #137 in the course planner[1], I wound up seeing a few dozen test failures that seemed to be related to the closing behavior of the modal. It seems that there were two issues:

1. Removing the `evt.preventDefault()` call on the Modal itself was causing some click events to register on the backdrop, closing the modal before it should have.
2. Defining the `mouseDownOnModal` field in the render function was causing it to revert to `false` when the component re-rendered, preventing the modal from closing when it should have.

At least, I _think_ that's what was happening. I don't think either of those were really registering as issue when manually testing the browser, but I suspect something in the testing library click event handling was causing the unexpected behavior. 

---

[1]: I tested this by first transpiling the code in `mark-one` with:

```
tsc --project=tsconfig.publish.json
```

Then copying the built files into my copy of course-planner:

```
cp -r /path/to/mark-one/lib /path/to/course-planner/node-modules/mark-one/lib
```

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #___

### DIFF
--- a/src/Modals/Modal.tsx
+++ b/src/Modals/Modal.tsx
@@ -199,6 +199,7 @@ const Modal: FunctionComponent<ModalProps> = ({
             onMouseDown={(): void => {
               mouseDownOnModal = true;
             }}
+            onClick={(evt): void => { evt.stopPropagation(); }}
             theme={theme}
             ref={finalForwardRef}
           >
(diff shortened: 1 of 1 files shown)


natalynyu (approved):
Looks good to me. Approved!

PR #141, opened 2023-01-10
TableHeadingCell updated adding padding.

Urjikoo:
## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #602


jonseitz (changes requested):
So this will add padding to the `TableHeadingCell`, which is good, but for the case we're trying to address in the courses page we actually need to update the `TableRowHeadingCell` as well, as that's what's being [used for the Course column on the page](https://github.com/seas-computing/course-planner/blob/main/src/client/components/pages/Courses/CourseInstanceTableBody.tsx#L78-L86).

(Sorry, I had forgotten those were two separate components when filling out the original ticket)

jonseitz (approved):
This is all set! Nice job.

Once this is merged we can merge develop into main here, then bump the version of mark-one used in Course Planner

PR #152, opened 2023-02-21
Feature/create component for table cell to stick

Urjikoo:
## Describe your changes
This PR addresses issue #345 on course planner to implement TableHead position when the page/table is scrolled. New prop called isStick created to make TableHead component on the tables to stick during scrolling. To make sure the prop works, i've added it to TableHead in the table test data.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #

### DIFF
--- a/src/Tables/Table.tsx
+++ b/src/Tables/Table.tsx
@@ -2,7 +2,6 @@ import styled from 'styled-components';
 import { ReactElement, ForwardRefExoticComponent } from 'react';
 import TableHead from './TableHead';
 import TableBody from './TableBody';
-import { fromTheme } from '../Theme';
 
 export interface TableProps {
   /**


jonseitz (changes requested):
This is really great start, but I think there are still some rough edges that need to be fixed before this is ready. I pulled these changes into my local course planner repository with:

```sh
$ cd mark-one
$ rm -rf ./lib/
$ npm run prepublishOnly
$ rm -rf ../course-planner/node_modules/mark-one/lib/
$ cp ./lib ../course-planner/node_modules/mark-one/
```
_Assumes that `course-planner` and `mark-one` are sibling directories_

After restarting the course-planner container and setting `isSticky` on the Courses table, I get the following issues:

## Firefox

Apparently there's [a bug][bug] where it won't display the cell borders when you have a sticky `<thead>` inside of a table with `border-collapse: collapse` set. So our multi-level header loses a lot of readability.


 
## Chromium

Blink-based browsers don't seem to have the same problem as Firefox; however, I'm still seeing the issue where the background content shows through the borders while scrolling the table:



I did do a little investigation and I've seen a few places where people recommend setting the sticky property on the <th> elements instead of the `<thead>`, though I haven't tested that myself. 

I'll keep poking around on this and see if I can find any other solutions, but we need to get these bugs resolved before this gets merged.

[bug]: https://bugzilla.mozilla.org/show_bug.cgi?id=1727594

examples/Tables/Table.md
```diff
@@ -171,7 +171,7 @@ import {
 } from 'mark-one';
 
 <Table>
-  <TableHead>
+  <TableHead isSticky={true}>
```
jonseitz, line 174: For a more advanced feature like this, I think it would be good to break out a separate section of the documentation that explains a little more about the situations where you would want to use `isSticky`, rather than just add it into one of the other examples. You could also create an example of a very long table to show how it works in practice (You can always an array iterator to generate a bunch of rows automatically).

jonseitz (changes requested):
Nice job getting the browser compatibility issues sorted out! Unfortunately fixing that introduced some other little visual quirks that I've noted here. I think getting the borders to behave is probably the most intensive. I've included an example using some nested CSS styles that target specific <td> and <th> elements and apply borders individually. I'm afraid that might be the only way to get the styling to look as it's intended. I tried to leave comments about what each layer is doing, but please feel free to let me know if you want to chat in more detail about any of it.


On a more general note for the future, I'd like you to spend a little more time on your commit messages. In particular there are a couple of commits here that have overly long subject lines, which get truncated by github. In general, subject lines should be about 50 characters, and should briefly list what change the commit is introducing. The body of the message should clarify _why_ that change is needed, and introduce any additional context about the process that lead there, e.g links to documentation, any previous failed attempts, etc. "Context" is really the key here. The goal is provide additional information that's not apparent from the code itself, so that some day when we're trying to debug an issue we can get a little more insight into why something was done the way it was. 

Personally, I've found this "[Seven Rules][seven]" guide to be really helpful.  

[seven]: https://cbea.ms/git-commit/#seven-rules

examples/Tables/Table.md
```diff
@@ -202,6 +202,76 @@ import {
...
+    <TableRow>
+      <TableCell>8</TableCell>
+      <TableCell>Glenn</TableCell>
+      <TableCell>Kristin</TableCell>
+    </TableRow>
+    <TableRow>
+      <TableCell>9</TableCell>
+      <TableCell>Gabriela</TableCell>
+      <TableCell>Hines</TableCell>
+    </TableRow>
```
jonseitz, line 271: Thanks for adding this extended example. I think it's worth generating *a lot* of rows (i.e. in the hundreds) for this, to really show where the feature is useful. We don't need the data to be particularly meaningful, so it's fine to just fill with a sequence of numbers in different columns.

Obviously we don't want to create those by hand, but we can use the second argument to  [`Array.from()`][mdn] method (which is essentially the same as `Array.map()`) to generate them for us, e.g.:

```ts
// To generate a list from 0 to 9
<ol>
  {Array.from({length: 10}, (_, index) => (
    <li>{index}</li>)
  )}
</ol>
```

We should also make sure to extend the row striping all the way down. It may also be helpful to create a simple wrapper component around the table, with absolute positioning, scrolling overflow, and a fixed value for the height. (You can always import the `styled` tool here and define a separate `styled.div` directly in the example body that we can use as a wrapper).

Basically, the goal is just to make sure that our style guide resembles our real use cases as much as possible, both as a reminder for the future and to help us catch any issues in the process. Let me know if you need any help getting that set up!

jonseitz (changes requested):
A few more notes below, including a change from last round that wasn't addressed. It might be helpful to go through the comments and respond to each with a reference to a related commit that addresses the issue, if only as a checklist of sorts.

src/Tables/Table.tsx
```diff
@@ -2,6 +2,7 @@ import styled from 'styled-components';
 import { ReactElement, ForwardRefExoticComponent } from 'react';
 import TableHead from './TableHead';
 import TableBody from './TableBody';
+import { fromTheme } from '../Theme';
```
jonseitz, line 5: Last little thing: This import of `fromTheme` should be removed if it's not being used.

jonseitz (changes requested):
One small note about removing an unused import, but once that's done this is all set!

jonseitz (approved):
This is all set!

PR #155, opened 2023-05-02
Bugfix 153/footer should stick to the bottom of the page

### LINKED ISSUE #153: Footer should stick to the bottom of the page
When initially loading one of our large tables, the footer will appear immediately below the Loading Spinner as that is the only content on the page (as in the screenshot below). It feels a little weird to have the footer in the middle of the page, so I propose we set the CSS `position` value such that the footer will always appear at the bottom of the page, regardless of how long the content is.

2023-04-12_143819_crop

Urjikoo:
## Describe your changes
 The footer component on course planner page did not have a fixed position causing it to move way up during  the page gets refreshed and also when the table size get small. To fix that, position added and solid white(light) background color added to make it visible and not blend with the table. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #153

### DIFF
--- a/src/Layout/Footer.tsx
+++ b/src/Layout/Footer.tsx
@@ -29,8 +29,8 @@ const Footer = styled.footer<FooterProps>`
   position: fixed;
   bottom: 0;
   height: 40px;
-  font-size: 0.9em;
-  font-weight: 400;
+  font-size: ${({ theme }) => `${theme.font.base.size}`};
+  font-weight: ${({ theme }) => `${theme.font.base.weight}`};
 `;
 
 Footer.defaultProps = {


src/Layout/Footer.tsx
```diff
@@ -21,11 +21,17 @@ export interface FooterProps {
...
   align-items: baseline;
-  background-color: ${({ background }) => background};
+  background: ${({ theme }) => theme.color.background.light};
   display: flex;
   justify-content: ${({ justify }) => justify};
   padding: ${({ theme }) => `${theme.ws.medium} ${theme.ws.small}`};
   width: 100%;
+  position: fixed;
+  bottom: 0;
+  width: 100%;
```
rmainseas, line 31: This property is already set on line 28, so we don't need it here also.
```suggestion
```

src/Layout/Footer.tsx
```diff
@@ -21,11 +21,17 @@ export interface FooterProps {
...
+  height: 40px;
+  font-size: 0.9em;
+  font-weight: 400;
```
rmainseas, line 34: Rather than hard code this, I'd suggest using the `theme` values here instead. Feel free to tweak these as needed to whatever you feel is more appropriate (i.e: `theme.font.bold.size` etc. - vs code should be able to typehint the available options for you)

```suggestion
  font-size: ${({ theme }) => `${theme.font.base.size}`};
  font-weight: ${({ theme }) => `${theme.font.base.weight}`};
```

rmainseas (changes requested):
Looks great, just a few small notes.

rmainseas (approved):
:rocket: Looks good to me!

PR #158, opened 2023-10-16
Fix PageBody

natalynyu:
The purpose of adding a padding to `PageBody` is to solve for an issue in Course Planner in which the `Footer` covers the bottom of a table making the information in the table inaccessible to the user. By adding some padding to the bottom of `PageBody`, we are making it so that the footer does not obfuscate any information.

The purpose of adding a padding to PageBody is to solve for an issue in Course Planner in which the footer covers the bottom of a table making the information in the table inaccessible to the user. By adding some padding to the bottom of PageBody, we are making it so that the footer does not obfuscate any information.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #704

### DIFF
--- a/src/Layout/PageBody.tsx
+++ b/src/Layout/PageBody.tsx
@@ -5,11 +5,13 @@ import { fromTheme } from '../Theme';
 export interface PageBodyProps {
   /** Page contents */
   children?: ReactNode;
+  /** Bottom padding of the page body */
+  paddingBottom?: string;
 }
 
 const PageBody = styled.main<PageBodyProps>`
   margin: ${fromTheme('ws', 'small')};
-  padding-bottom: 40px;
+  padding-bottom: ${({ paddingBottom, theme }): string => (paddingBottom || theme.ws.xlarge)};
 `;
 
 declare type PageBody = ReactElement<PageBodyProps>;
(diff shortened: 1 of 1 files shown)


src/Layout/PageBody.tsx
```diff
@@ -9,6 +9,7 @@ export interface PageBodyProps {
 
 const PageBody = styled.main<PageBodyProps>`
   margin: ${fromTheme('ws', 'small')};
+  padding-bottom: 40px;
```
robertmain, line 12: I've had a look at this on info.seas.harvard.edu and it seems as though the height of the footer is obscuring the bottom of the `PageBody` component
image

Given that, setting `padding: 40px;` on the `PageBody` component as you have here does indeed fix the issue

image

However, if the footer is made taller, we're likely going to run into the same issue again.

image

Given that - I wonder if it makes sense to instead have either have optional `padding` parameter(s)? that one can pass(as well as a `height` parameter we can pass to [`Footer.tsx`](https://github.com/seas-computing/mark-one/blob/develop/src/Layout/Footer.tsx#L31) at some point in the future)

robertmain (changes requested):
Thanks for doing this, Natalyn! Just a note regarding hard coding the padding value

src/Layout/PageBody.tsx
natalynyu: Thanks Robert, that's a really good point. I updated the `PageBody` component to have an optional parameter for `padding` to give the user more flexibility.

rmainseas: Looks great. Merge at will.

rmainseas (approved):
Looks great! :rocket:

PR #163, opened 2023-12-08
Allow for non-floating footer

### LINKED ISSUE #161: Unfloat the application footer
As Vittorio noted in [the accessibility review on the updated apply.seas](https://docs.google.com/document/d/1RtjswP0AsOYeaoAsKx5_ob1e60Cox_aF2QNhB0U5KEk/edit#heading=h.8hy5063nh75l) we don't necessarily need the footer to float on this page, and doing so is causing text to disappear at smaller screen width. I think we may want to make the floating aspect optional, or otherwise provide some special handling for display of the footer at smaller widths, i.e. stacking the fields

jonseitz:
## Describe your changes

This adds a `position` prop to the `<Footer>` component, which is passed through the the CSS property of same name to allow for customization of the Footer styling. It also changes the default from `fixed` to `sticky`, which means that it will "un-float" when the window scrolls to the end of the page content -- that should fix the issue we've sometimes seen with the footer covering up the bottom of the page.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #161

### DIFF
--- a/examples/Layout/Footer.md
+++ b/examples/Layout/Footer.md
@@ -24,6 +24,19 @@ You can also pass in a custom `background` prop to replace the default transpare
 </Footer>
 ```
 
+The `position` prop can be used to control how the footer is positioned on the
+page. By default, the footer will be `sticky`, appearing `fixed` while
+scrolling the page, then `relative` when it reaches the bottom of the content.
+You can set it to always appear at the end of the content by setting position
+to `relative`."
+
+
+```jsx
+<Footer position="relative">
+  <div>Version 10.0</div>
+</Footer>
+```
+
 An unordered list can be provided to the `Footer`, and each list item will be spaced accordingly.
 ```jsx
 import { ExternalLink } from 'mark-one';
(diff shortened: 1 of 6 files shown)


robertmain (approved):

rmainwork (approved):
This looks great. I especially love the typed CSS properties, I didn't know that was a thing!

Also, I'm curious as to how you tested this - did you have to downgrade your version of node for the time being? I can't get styleguidist server to run properly on Node 18.

jonseitz:
> Also, I'm curious as to how you tested this - did you have to downgrade your version of node for the time being? I can't get styleguidist server to run properly on Node 18.

Sort of? I just ran it in a `node:14` docker container and was able to get it working:

```sh
docker run -it --rm -p 127.0.0.1:6060:6060 -v $(pwd):/node node:14 bash -c "cd /node; npm run start"
```

I did also play around with upgrading `webpack`, `styleguidist`, etc for a bit and got it to compile, but any time I tried to load the styleguide there was an error with some kind of polyfill it was trying to apply.

PR #165, opened 2024-01-03
Fix export of theme type

### LINKED ISSUE #162: Fix export of `MarkOneTheme` type
We've defined a `MarkOneTheme` type that should override the existing `DefaultTheme` type in consuming applications. However, it's not currently being exported correctly since it's created as a `.d.ts` file. So we've wound up copying the file into our other project so we can access those types. We should make that a regular `.ts` file so that it is compiled and distributed with the rest of `mark-one`.

jonseitz:
## Describe your changes

It seems that the `DefaultTheme` override in the `styled.d.ts` file was causing problems in the build phase -- mainly that the declarations were not being carried over into the final output. To resolve, we're removing the `d.` from the filename, and exporting an empty module from the new file. Importing that empty module into our index barrel file seems to be enough to pick up the declarations and package them into the final output.

I tested this by running `npm run prepublishOnly` in the mark-one project root, which builds the code under `lib/`. I then copied that entire `lib` folder into the `node_modules/mark-one/` directory of a different project (i.e. `apply-seas-node`), removed the `styled.d.ts` from that project, and confirmed that typescript is still able to pick up the typing for added keys like `ws.small`.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #162

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,5 @@
+import './styled';
+
 export * from './Buttons';
 export * from './Layout';
 export * from './Tabs';
--- a/src/styled.ts
+++ b/src/styled.ts
@@ -14,7 +14,7 @@ import {
 declare module 'styled-components' {
   export interface DefaultTheme {
     color: {
-      background:{
+      background: {
         [K in VARIANT | keyof ColorRange]:
         K extends VARIANT
           ? ColorRange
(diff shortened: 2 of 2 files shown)


natalynyu (approved):
Looks good to me! I went through each of the steps for testing and it worked.

PR #173, opened 2024-03-20
Fix Form and Button components

natalynyu:
This PR updates the `Form` component to fix the following error that we are receiving when trying to use it: 
```
Type '{ children: Element[]; id: string; label: string; submitHandler: () => void; }' is not assignable to type 'IntrinsicAttributes & FormProps'.
  Property 'children' does not exist on type 'IntrinsicAttributes & FormProps'. 
```
We have seen this error in the past when trying to update Mark One to React 18. This was solved by explicitly passing down `children` as a prop.

The PR also updates the `Button` component to include an optional `type` property. If not specified, the `type` defaults to `button`. This way, we will be able to specify a `type` of `submit` so that we can more properly handle and associate forms with submit buttons.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes [#40](https://github.huit.harvard.edu/SEAS/makerspace/issues/40)

### DIFF
--- a/src/Buttons/Button.tsx
+++ b/src/Buttons/Button.tsx
@@ -5,6 +5,7 @@ import React, {
   ReactNode,
   MouseEventHandler,
   Ref,
+  ButtonHTMLAttributes,
 } from 'react';
 import styled, { ThemeContext } from 'styled-components';
 import {
@@ -31,7 +32,7 @@ export interface ButtonProps extends MarkOneProps<HTMLButtonElement> {
   /** Specifies the ref of the element */
   forwardRef?: Ref<HTMLButtonElement>;
   /** Specifies the button type */
-  type?: string;
+  type?: ButtonHTMLAttributes<HTMLButtonElement>['type'];
 }
 
 const StyledButton = styled.button<ButtonProps>`
--- a/src/Buttons/Button.tsx
+++ b/src/Buttons/Button.tsx
@@ -30,6 +30,8 @@ export interface ButtonProps extends MarkOneProps<HTMLButtonElement> {
   form?: string;
   /** Specifies the ref of the element */
   forwardRef?: Ref<HTMLButtonElement>;
+  /** Specifies the button type */
+  type?: string;
 }
 
 const StyledButton = styled.button<ButtonProps>`
@@ -66,6 +68,7 @@ const Button: FunctionComponent<ButtonProps> = (props): ReactElement => {
     alt,
     className,
     form,
+    type,
   } = props;
   const theme = useContext(ThemeContext);
   return (
@@ -79,6 +82,7 @@ const Button: FunctionComponent<ButtonProps> = (props): ReactElement => {
       aria-label={alt}
       className={className}
       form={form}
+      type={type}
     >
       { children }
     </StyledButton>
@@ -91,6 +95,7 @@ Button.defaultProps = {
   disabled: false,
   form: '',
   forwardRef: null,
+  type: 'button',
 };
 
 export default Button;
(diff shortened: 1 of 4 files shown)


src/Forms/Form.tsx
```diff
@@ -1,33 +1,51 @@
...
-  })
-)`
-  & > * {
-    margin: ${fromTheme('ws', 'small')};
-  }
+const StyledForm = styled.form`
+  margin: ${fromTheme('ws', 'small')};
 `;
 
+const Form: FunctionComponent<FormProps> = (props): ReactElement => {
```
rmainseas, line 26: We can actually use `PropsWithChildren` here so that we don't have to type `children` directly in the `props` object

```suggestion
const Form: FunctionComponent<PropsWithChildren<FormProps>> = (props): ReactElement => {
```

src/Buttons/Button.tsx
```diff
@@ -30,6 +30,8 @@ export interface ButtonProps extends MarkOneProps<HTMLButtonElement> {
   form?: string;
   /** Specifies the ref of the element */
   forwardRef?: Ref<HTMLButtonElement>;
+  /** Specifies the button type */
+  type?: string;
```
rmainseas, line 34: What do you think about using `ButtonHTMLAttributes` to get better type safety here?


```suggestion
  type?: ButtonHTMLAttributes<HTMLButtonElement>['type'];
```

src/Forms/__tests__/Form.test.tsx
```diff
@@ -18,7 +21,22 @@ describe('Form', function () {
...
+          onChange={onChangeStub}
+        />
+        <Button
+          id="testButton"
+          form="testForm"
+          type="submit"
+          variant={VARIANT.PRIMARY}
+          onClick={submitStub}
+        />
+      </Form>
```
rmainseas, line 39: Are these serving a purpose in this test?

rmainseas (changes requested):
Just a few small notes, but this is looking great - thanks for working on this!

src/Forms/__tests__/Form.test.tsx
natalynyu: I didn't think that we could just have a standalone `Form` element without any field inputs/submit button inside of it. Is it okay on its own?

src/Forms/Form.tsx
natalynyu: Ah sorry I had forgotten about this possibility. Thanks for pointing this out! (Fixed in [3f87a74](https://github.com/seas-computing/mark-one/pull/173/commits/3f87a742c75969f5fc9901fd5fdaf674c0d0bf7c))

src/Buttons/Button.tsx
natalynyu: That's a great point. I updated the typing in [3dac3f2](https://github.com/seas-computing/mark-one/pull/173/commits/3dac3f24ed30414ea27c2639c590cf4a3a64adaf).

src/Forms/__tests__/Form.test.tsx
rmainseas: In general, I'd say that a form doesn't make sense on it's own(certainly for the documentation), but since this is a unit test, I think it might be better to keep the test tightly scoped to what we're actually trying to test. Since `Button` and `TextInput` aren't technically part of this test, it might be best to remove them.

src/Forms/Form.tsx
rmainseas: No worries!

rmainseas (changes requested):
This is looking great - just a quick note about the test

rmainseas (approved):
Approved - good to go!

PR #189, opened 2024-08-13
Add Callout Component

### LINKED ISSUE #185: Callout
The `Callout` is intended to display a standalone piece of text that should stand out to the user. In the "shops.seas" project, this would be used to display reminders to users about taking required safety trainings. 

It should be a block-level element (i.e. `<p>`), its text size should be larger than our standard type, it should have extra margins to separate it from the rest of the text.

amalyam:
This PR adds a `Callout` component, which is intended to display a piece of text that should stand out to the user. It is styled on a `<p>` element, the text size is larger than the standard type, and it has extra margins to separate it from the rest of the text.

You can pass a variant property from the `VARIANT` enum to set the background color. For accessibility purposes, only the `light` option from each variant's `ColorRange` is used, and the text color will always be `dark`. Additionally, the `role` property can be set to define the ARIA role of the callout, with a default value of 'note'.

## Types of changes
      to change)

## Checklist:

## Priority:
- [x] Normal
- [ ] High

## Related Issues:
Closes #185

### DIFF
--- a/examples/Layout/Callout.md
+++ b/examples/Layout/Callout.md
@@ -1,0 +1,5 @@
+Renders a styled `<p>` element for displaying a piece of text intended to stand out to the user.
+
+```jsx
+<Callout>Callout text</Callout>
+```
--- a/examples/Layout/Callout.md
+++ b/examples/Layout/Callout.md
@@ -1,5 +1,25 @@
 Renders a styled `<p>` element for displaying a piece of text intended to stand out to the user.
 
+The following is an example of the default callout in which the `variant` property is assigned the default enum value `VARIANT.INFO` and the role is assigned the default value `note`.
+
 ```jsx
-<Callout>Callout text</Callout>
+import { VARIANT } from 'mark-one';
+
+<Callout>You currently have 3 shop safety certifications.</Callout>
+```
+
+The following is an example of a callout in which the `variant` property is assigned the enum value `VARIANT.POSITIVE` and the role is assigned the default value `note`.
+
+```jsx
+import { VARIANT } from 'mark-one';
+
+<Callout variant={VARIANT.POSITIVE}>Your submission has been received! Please check your inbox for a confirmation email. We will respond within the next 48 hours.</Callout>
+```
+
+The following is an example of a callout in which the `variant` property is assigned the enum value `VARIANT.NEGATIVE` and the role is assigned the value `alert`.
+
+```jsx
+import { VARIANT } from 'mark-one';
+
+<Callout variant={VARIANT.NEGATIVE} role='alert'>You must attend a lab safety training before taking this course.</Callout>
 ```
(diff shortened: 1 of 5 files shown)


src/Layout/Callout.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * @default VARIANT.INFO
+ */
+  variant?: VARIANT;
+  /**
+   * The aria role of the displayed callout. A list of the different role
+   * values that should be used can be found in the w3 docs:
+   * https://www.w3.org/TR/wai-aria-1.1/#live_region_roles
+   * The default value role is 'note'.
+   */
+  role?: string;
```
jonseitz, line 17: There's [an `AriaRole` type][1] defined in the `@types/react` package, which we could use here to ensure that the passed value is actually a valid role.

```ts
import React, { AriaRole } from 'react';

export interface CalloutProps {
  role?: AriaRole;
}
```

[1]: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/298040491bdf086e6fbfe3c9eb12fbe96c90b7fc/types/react/index.d.ts#L2815-L2885

src/Layout/Callout.tsx
```diff
@@ -0,0 +1,57 @@
...
+  variant: VARIANT;
+}
+
+const StyledCallout = styled.p<StyledCalloutProps>`
+  font-family: ${({ theme }): string => theme.font.callout.family};
+  font-size: ${({ theme }): string => theme.font.callout.size};
+  font-weight: ${({ theme }): string => theme.font.callout.weight};
+  color: ${({ theme }): string => theme.color.text.dark};
+  background: ${({ theme, variant }): string => theme.color.background[variant].light};
+  margin-left: ${({ theme }):string => theme.ws.large};
```
jonseitz, line 31: I would recommend duplicating this for `margin-right` as well, for cases where a `Callout` might fill the entire horizontal space. Alternatively, we could define this as a single margin prop, along the lines of:

```
  margin: ${({ theme }):string => `${theme.ws.medium} ${theme.ws.large} ${theme.ws.small} ${theme.ws.large}`};
```

examples/Layout/Callout.md
```diff
@@ -0,0 +1,5 @@
+Renders a styled `<p>` element for displaying a piece of text intended to stand out to the user.
+
+```jsx
+<Callout>Callout text</Callout>
+```
```
jonseitz, line 5: For components whose styling can be changed via props, it's usually good to provide a few examples showing some of the different use cases. So here, we might want to have:

1. `VARIANT.INFO` callout
2. `VARIANT.NEGATIVE` callout (maybe in this case also setting `role: 'alert'`)
3. `VARIANT.POSITIVE` callout

For each, I'd also suggest using text that might plausibly be used in an application, like the "Remember to attend the lab safety training." message we have on the sign-in kiosk. Especially for a block-level component like this, it's good to show how it might look with different amounts of text.

jonseitz (changes requested):
This is looking really good! Just a couple little things noted below that need to be fixed. Let me know if you want to chat about any of these further.

amalyam:
@jonseitz Thank you for the notes and tips!

jonseitz (approved):
This looks great! You may merge when ready!

PR #192, opened 2024-08-22
Add GridWrapper Component

### LINKED ISSUE #181: Layouts: GridWrapper
The `GridWrapper` will be the top-level of our grid structure. It's only purpose should be to define the 12-column grid, with space between rows and columns. It should not apply any margins or padding, and should not have any additional borders shadows or other styling. Those non-grid aspects can be handled by parent/child components instead.

In use, we'll be able to embed grids within other components, letting us define sub-grid arrangements when needed.

amalyam:
## Describe your changes
This PR adds a `GridWrapper` component which serves as the top-level of the grid structure. It defines the 12-column grid, with space between rows and columns. It can be used to embed grids within other components, letting us define sub-grid arrangements when needed.

It provides default gap spacing and responsive gap spacing for small window sizes, both of which can be overridden by passing in a gap size using the `MarkOneTheme` `ws` values. The component does not apply any margins, padding, additional borders shadows, or other styling, which should be handled by parent/child components instead.

This PR also adds a GridWrapper markdown example, which will be replaced with a more thorough example using the `GridContainer` and `GridItem` components, once all three are complete.

## Types of changes
      to change)

## Checklist:

## Priority:
- [x] Normal
- [ ] High

## Related Issues:
Closes #181

### DIFF
--- a/examples/Layout/GridWrapper.md
+++ b/examples/Layout/GridWrapper.md
@@ -1,0 +1,30 @@
+*Note: this example is a placeholder until the `GridContainer` and `GridItem` components are finished.
+
+The `GridWrapper` is the top-level of the grid structure. It defines the 12-column grid, with space between rows and columns. It can be used to embed grids within other components, letting us define sub-grid arrangements when needed.
+
+It does not apply any margins, padding, additional borders shadows, or other styling, which should be handled by parent/child components instead.
+
+```jsx
+    <div>
+    <h2>Small Gap (default)</h2>
+      <GridWrapper gap="small">
+        <div style={{ backgroundColor: 'lightblue', padding: '10px' }}>Item 1</div>
+        <div style={{ backgroundColor: 'lightgreen', padding: '10px' }}>Item 2</div>
+        <div style={{ backgroundColor: 'lightcoral', padding: '10px' }}>Item 3</div>
+      </GridWrapper>
+      <br/>
+      <h2>Medium Gap</h2>
+      <GridWrapper>
+        <div style={{ backgroundColor: 'lightblue', padding: '10px' }}>Item 1</div>
+        <div style={{ backgroundColor: 'lightgreen', padding: '10px' }}>Item 2</div>
+        <div style={{ backgroundColor: 'lightcoral', padding: '10px' }}>Item 3</div>
+      </GridWrapper>
+      <br/>
+      <h2>Large Gap</h2>
+      <GridWrapper gap="large">
+        <div style={{ backgroundColor: 'lightblue', padding: '10px' }}>Item 1</div>
+        <div style={{ backgroundColor: 'lightgreen', padding: '10px' }}>Item 2</div>
+        <div style={{ backgroundColor: 'lightcoral', padding: '10px' }}>Item 3</div>
+      </GridWrapper>
+      </div>
+```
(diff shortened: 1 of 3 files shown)


examples/Layout/GridWrapper.md
```diff
@@ -0,0 +1,30 @@
...
+    <div>
+    <h2>Small Gap (default)</h2>
+      <GridWrapper gap="small">
+        <div style={{ backgroundColor: 'lightblue', padding: '10px' }}>Item 1</div>
+        <div style={{ backgroundColor: 'lightgreen', padding: '10px' }}>Item 2</div>
+        <div style={{ backgroundColor: 'lightcoral', padding: '10px' }}>Item 3</div>
+      </GridWrapper>
+      <br/>
+      <h2>Medium Gap</h2>
+      <GridWrapper>
```
jonseitz, line 17: I think this example is missing a `gap="medium"` prop

examples/Layout/GridWrapper.md
```diff
@@ -0,0 +1,30 @@
+*Note: this example is a placeholder until the `GridContainer` and `GridItem` components are finished.
+
+The `GridWrapper` is the top-level of the grid structure. It defines the 12-column grid, with space between rows and columns. It can be used to embed grids within other components, letting us define sub-grid arrangements when needed.
+
+It does not apply any margins, padding, additional borders shadows, or other styling, which should be handled by parent/child components instead.
+
+```jsx
+    <div>
+    <h2>Small Gap (default)</h2>
+      <GridWrapper gap="small">
```
jonseitz, line 10: Since this is meant to show the default, maybe we can leave out the `gap="small"`? Or break it into two examples showing "no value" and "small"?

src/Layout/GridWrapper.tsx
```diff
@@ -0,0 +1,39 @@
...
+  children: ReactNode;
+}
+
+const StyledGridWrapper = styled.div<GridWrapperProps>`
+  display: grid;
+  grid-template-columns: repeat(12, 1fr);
+  gap: ${({ theme, gap }) => theme.ws[gap || 'small']};
+
+  @media (max-width: 768px) {
+    gap: ${({ theme, gap }) => theme.ws[gap || 'xsmall']}; /* Smaller gap for mobile devices */
```
jonseitz, line 21: It looks like the `defaultProps` value below is breaking this conditional -- the value of `gap` will always be defined so it will never fall through to the `'xsmall'` option at smaller sizes. Maybe we can just change the default value to `null`?

jonseitz (changes requested):
This looks great! I especially like how you used of `keyof` to limit the types available for the `gap` prop. I do have just a couple of minor implementation details to fix up, noted below.

examples/Layout/GridWrapper.md
amalyam: Forgot to change this and the previous issue after moving some things around, thanks for catching these. I'm planning to make a more robust example (like the Table example we talked about) after I finish all 3 components, so this is mostly just a placeholder so I can see how the component is functioning as I'm working on it.

src/Layout/GridWrapper.tsx
amalyam: Makes sense, null seems like the better way to go.

jonseitz (approved):
This looks like it's all set!

PR #198, opened 2024-09-17
Toast Notification Wrapper

### LINKED ISSUE #197: Toast Notification Wrapper Component
This component will wrap the `ToastNotification` component(s) such that they are aligned vertically at the top right of the browser.  We will determine in the future whether we will use a scroll bar for notifications that don't fit on the page or add a button to show additional notifications.

natalynyu:
The `ToastNotificationWrapper` component will be used to position the `ToastNotification` components such that they line up at the top right of the page.

The example includes a button to show the `ToastNotification` components wrapped in the `ToastNotificationWrapper` at the top right of the page. You will most likely need to scroll up to see the example, as it is purposely positioned outside of the regular example container. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #197

### DIFF
--- a/src/Alerts/ToastNotificationWrapper.tsx
+++ b/src/Alerts/ToastNotificationWrapper.tsx
@@ -6,13 +6,15 @@ import React, {
 } from 'react';
 import styled from 'styled-components';
 import { Button } from '../Buttons';
-import { VARIANT } from '../Theme';
+import { VARIANT, fromTheme } from '../Theme';
 
 export interface ToastNotificationWrapperProps {
   /** Elements to be displayed within the wrapper */
   children: ReactNode;
   /** Function to call on click event */
   onClick: MouseEventHandler;
+  /** Indicates how many notifications there are within the wrapper */
+  numNotifications?: number;
 }
 
 const StyledToastNotificationWrapper = styled.div`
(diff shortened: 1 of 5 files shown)


src/Alerts/ToastNotificationWrapper.tsx
```diff
@@ -0,0 +1,18 @@
...
+const ToastNotificationWrapper = styled.div`
+  display: flex;
+  align-self: flex-end;
+  flex-direction: column;
+  height: 100%;
+  overflow-y: auto;
+  position: absolute;
+  top: 5px;
+  right: 5px;
+`;
```
amalyam, line 16:
```suggestion
z-index: 1000;
`;
```
I think we also need to set a high z-index for accessibility and usability (to ensure it remains on top of other content)

amalyam (changes requested):
This looks good! In addition two the two inline comments I left, I had two other thoughts: 
 - should we have a prop like "autoHideDuration" so we can set an interval after which these will automatically close?
 - I was looking at MUI's [toast component](https://mui.com/material-ui/react-snackbar/) - they provide an "anchorOrigin prop to control the Snackbar's position on the screen". May not be necessary for now, but might be nice to include more than one option in case the default covers important content?

src/Alerts/ToastNotificationWrapper.tsx
```diff
@@ -0,0 +1,18 @@
...
+ * A wrapper for the ToastNotification component to position them at the top
+ * right of the page.
+ */
+const ToastNotificationWrapper = styled.div`
+  display: flex;
+  align-self: flex-end;
+  flex-direction: column;
+  height: 100%;
+  overflow-y: auto;
+  position: absolute;
```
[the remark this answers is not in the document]
natalynyu: Yes, that's an excellent idea! I updated the component to use `position: fixed` instead in commit [f4cb945](https://github.com/seas-computing/mark-one/pull/198/commits/f4cb945a5604d9658993c8f9b36ec34a6e46013a).

src/Alerts/ToastNotificationWrapper.tsx
natalynyu: Good point. This was updated in commit [99f6377](https://github.com/seas-computing/mark-one/pull/198/commits/99f6377d6206e410518ae2e0f048ef8e09ea07e0).

natalynyu:
Thank you for the suggestions, @amalyam! We've discussed in the past the possibility of having notifications disappear automatically, but we had some accessibility concerns. Vittorio had advised that having content disappear on its own could be tricky, since we'd have to make a decision on how long is long enough to display the message.

amalyam (approved):
Everything looks great and functions as expected! Nice work incorporating all of the different suggestions!

PR #205, opened 2024-10-02
Don't check test coverage

### LINKED ISSUE #204: Remove minimum test coverage threshold
Currently our CI tests are failing because test coverage isn't meeting the 80% threshold. This is not a great way to measure code/test quality, so rather than writing unnecessary tests to hit the coverage mark we should just turn off the minimum coverage as we've done in other projects.

jonseitz:
## Describe your changes

This is still running the coverage report and giving us the output after the tests run, but we don't need to fail the test run if we don't meet the 80% threshold. Philosophically, test coverage is not the best indicator of code quality or test quality, and we'd rather leave it to our best judgement about what needs to be tested, rather than writing tests just to make the number green.

## Types of changes

      to change)

## Checklist:

- [ ] I have run `eslint` on the code

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #204

### DIFF
--- a/.github/workflows/01-test-suite.yml
+++ b/.github/workflows/01-test-suite.yml
@@ -18,4 +18,3 @@ jobs:
       - run: npm run typecheck
       - run: npm run lint
       - run: npm run test
-      - run: npm run codecov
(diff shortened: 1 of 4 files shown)


natalynyu (approved):
This looks good to me!

PR #209, opened 2025-01-29
Fix MessageContext typing/exports

jonseitz:
## Describe your changes

These changes are needed to address a couple of small things in our implementation of the `Message` component in makerspace:

1. The current definition of `MessageReducerState` requires us to have a `currentMessage` value, but when we initialize the reducer we intentionally use `null` so we can conditionally render the message only when there is actually a message to show. We need to change that field to allow for `null` and `undefined` to better accommodate the actual use. (This was not being flagged as an error before, because we didn't have the `strictNullChecks` option enabled in `tsconfig.json` -- that option is needed to get a few of the recommended `@typescript-eslint` rules working correctly).

2. While we currently export all of the types and other values from the `MessageContext.tsx` file, we aren't exporting them from the `index.ts` barrel file. That means that we have to import things like `messageReducer` from `mark-one/lib/Alerts/MessageContext`. That works, but it makes it harder to follow our dependencies, so we should switch to exporting all of the values from `index.ts` instead.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #___

### DIFF
--- a/src/Alerts/MessageContext.tsx
+++ b/src/Alerts/MessageContext.tsx
@@ -53,7 +53,7 @@ export class AppMessage {
  */
 export interface MessageReducerState {
   queue: AppMessage[];
-  currentMessage: AppMessage;
+  currentMessage?: AppMessage | null;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Urjikoo (approved):
looks good merge away

PR #223, opened 2025-08-14
Fix FlexWrapper

### LINKED ISSUE #213: Fix FlexWrapper Component
### Description

Our current implementation for the [FlexWrapper](https://github.com/seas-computing/mark-one/blob/develop/src/Layout/FlexWrapper.tsx) component needs to be fixed. We've set it up now as (condensed):
```
const StyledFlexWrapper = styled.div`
...
`

const FlexWrapper = () => (
  <StyledFlexWrapper>
    {children}
  </StyledFlexWrapper>
)
```

So the rules we're trying to add only get applied to the FlexWrapper itself, which is a kind of transient component that doesn't actually get rendered; instead we just get the underlying StyledFlexWrapper styles without any of these changes.

We should modify this component to just directly export the styled.div instead of adding the additional transient component.

### Additional Information

We should make sure to update our mark-one package in makerspace after making this fix.

Previously discussed [here](https://github.huit.harvard.edu/SEAS/makerspace/pull/201#discussion_r23892)

amalyam:
## Describe your changes

This PR modifies the `FlexWrapper` so that there is no transient "styled" layer - the FlexWrapper is exported directly. This should fix an issue where styles were not being applied properly to the FlexWrapper.

## Types of changes

      to change)

## Checklist:

- [X] I have added JSDoc for all of my code (where applicable)

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #213

### DIFF
--- a/src/Layout/FlexWrapper.tsx
+++ b/src/Layout/FlexWrapper.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react';
+import { ReactNode } from 'react';
 import styled, { DefaultTheme } from 'styled-components';
 import CSS from 'csstype';
 
--- a/src/Layout/FlexWrapper.tsx
+++ b/src/Layout/FlexWrapper.tsx
@@ -29,30 +29,17 @@ export interface FlexWrapperProps {
   children: ReactNode;
 }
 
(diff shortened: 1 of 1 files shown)


natalynyu (approved):
Approved!

PR #231, opened 2026-06-11
Fix Modal Focus

### LINKED ISSUE #229: Fix Modal Focus
### Description

When a modal is open, it currently allows the user to tab into the browser's URL bar. For accessibility purposes, when a modal is open, keyboard tabbing should be restricted to only the content within the modal. The user should NOT be able to tab into the URL bar if a modal is active.

### Acceptance Criteria

- [ ] When the modal component is active, the user should not be able to move focus outside of the modal

### Additional Information

[W3C Modal Dialog Example](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/)
[W3C Dialog (Modal) Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/), with more details about the focus when the dialog is closed.

amalyam:
## Describe your changes

This PR fixes an accessibility issue with our Modal component. Previously, tabbing focus was not constrained to the modal when it was open, which was allowing users to tab into the url bar. Now tabbing focus is kept within the modal. I made sure that the focus is in the modal as soon as it is open, instead of it remaining on the page behind the modal until the user interacts with elements within the modal.

I also made the modal responsible for returning focus to the last element that was focused before it was opened, so we don't have to set this up each time we use a modal in our apps, as this behavior should always be in place for accessibility.
## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #229

### DIFF
--- a/src/Modals/Modal.tsx
+++ b/src/Modals/Modal.tsx
@@ -117,6 +117,18 @@ const StyledModal = styled.div`
 `;
 
 /**
+ * Invisible element to manage keyboard tabbing.
+ * Placing FocusSentinels before and after the modal
+ * content keeps focus trapped within the modal.
+ */
+const FocusSentinel = styled.div`
+  width: 0;
+  height: 0;
+  overflow: hidden;
+  outline: none;
+`;
+
+/**
  * @component Modal
  * A dialog style popup that can be used to alert the user to new information,
  * or to display additional editing fields/content. While open, the background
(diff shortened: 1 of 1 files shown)


natalynyu (approved):
This looks great! This is a really clever solution. Would you mind adding some clarifying comments within the two `onFocus` values within the `FocusSentinel`s? I think this will help refresh our memory as to why we're getting the `focusableElements` and `lastFocusable`/`firstFocusable` values in the future. 

Also, this may be a good item to demo at one of our dev standups!

PR #233, opened 2026-07-06
Add id prop to ModalHeader

### LINKED ISSUE #230: Add `id` prop to ModalHeader
### Description

`ariaLabelledBy` is currently a required prop in the `Modal` component, which requires that there is some corresponding element in the modal that has an `id` with the same value. The `ModalHeader` makes the most sense for this, but `id` is not currently an accepted prop on the `ModalHeader`. We should update this to ensure that assistive technology is able to convey meaningful information to users.

### Additional Information

[MDN's guide to the aria-labelledby attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby)

---

- [ ] Test
- [ ] Docs

amalyam:
## Describe your changes

This PR adds a missing id prop to the ModalHeader that is needed in order to properly set the value provided by the `ariaLabelledBy` in the Modal component. It also fixes a semantic issue where the ref and tabIndex props were previously set on the StyledModalHeader (which is a div). They are now part of the ModalTitle (which is a SectionHeading).

This change ensures that assistive technology is able to convey meaningful information about our modals to users.

## Types of changes

      to change)

## Checklist:

- [x] I have updated the tests to cover my changes.

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #230

### DIFF
--- a/examples/Modals/ModalHeader.md
+++ b/examples/Modals/ModalHeader.md
@@ -10,6 +10,7 @@ import {
 } from 'mark-one';
 
 const [modalVisible, setModalVisible] = useState(false);
+const modalTitleId = "modal-header-example-1"
 
 <>
   <Button
@@ -19,11 +20,12 @@ const [modalVisible, setModalVisible] = useState(false);
     Open Modal
   </Button>
   <Modal
-    ariaLabelledBy="modalHeader-example-button-1"
+    ariaLabelledBy={modalTitleId}
     closeHandler={() => {setModalVisible(false)}}
     isVisible={modalVisible}
   >
     <ModalHeader
+      id={modalTitleId}
       closeButtonHandler={() => {setModalVisible(false)}}
     >
       Modal Header
@@ -68,6 +70,8 @@ const RefExample = () => {
     rendered by letting next task of event queue run first */
     setTimeout(() => ref.current.focus());
   }
+  const modalTitleId = "modal-header-example-2"
+
   return (
     <>
       <Button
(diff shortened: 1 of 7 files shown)


examples/Modals/ModalHeader.md
```diff
@@ -19,11 +20,12 @@ const [modalVisible, setModalVisible] = useState(false);
     Open Modal
   </Button>
   <Modal
-    ariaLabelledBy="modalHeader-example-button-1"
+    ariaLabelledBy={modalTitleId}
     closeHandler={() => {setModalVisible(false)}}
     isVisible={modalVisible}
   >
     <ModalHeader
+      id={modalTitleId}
```
Urjikoo, line 28: I couldn't comment on the modalHeader test, but since you've updated `ModalHeader`, the test should be updated with an `id`. ( you can run `npm run lint` to see where id is missing as well)

src/Modals/ModalHeader.tsx
```diff
@@ -17,6 +17,10 @@ interface ModalHeaderProps {
    * the Title to display in the Modal Header
    */
   children: ReactNode;
+  /**
+   * Takes the ariaLabelledBy prop in the Modal component
```
Urjikoo: Technically `ModalHeader`does not take an ariaLabelledBy prop, you can say ` id for the modalheader element referenced(to be targeted) by the parent Modal’s aria-labelledby` or you can word it however similar

Urjikoo (changes requested):
I had few suggestion regarding the test failures and docblock description other than that this is almost done. Let me know if you have any questions.

examples/Modals/ModalHeader.md
amalyam: Updated in [e88d591](https://github.com/seas-computing/mark-one/pull/233/commits/e88d59113344a8818021993ff7993d5eefbd7277)

src/Modals/ModalHeader.tsx
amalyam: Revised in [6ffb0a9](https://github.com/seas-computing/mark-one/pull/233/commits/6ffb0a94b95068158e59880db0fbab42a493fed4)

Urjikoo (approved):
Good to go! But before you merge, add the ticket number to the PR and why we had to make this update on mark one: i can see that it addresses semantic problem but if there is an error we're trying to fix with this on other projects like OAP, it'd be helpful to add it to the PR note

